PR #9, opened 2023-12-14
fix: Force sacct to look at the last 2 days

fgvieira:
Possible fix to snakemake/snakemake#1227.

How often does the rate limiter checks `sacct`?


cmeesters (approved):
looks okay, seems working

Just: Do not know whether a multitude of checks with 2d look back time might put some strain on some slurmdbds . This remains to be observed.

fgvieira:
We can also change it to 1 day, instead of 2.

cmeesters:
na, I'm fine with it. Just saying we need to pay a bit of attention, if there is any such feedback - ultimately, the user can limit the lookup rate, too.

PR #10, opened 2023-12-29
feat: print run id to log

fgvieira:
Print run `run_uuid` to log.


cmeesters:
Good idea! Yet, the job id is not the jobname (see comment to #16). The job id is `slurm_jobid` and just known after submission. And it is reported in lines 149-154 (current code).

fgvieira:
Not sure I follow... Here I am referring to the run uuid, not the job id or the slurm job id.

cmeesters (approved):
sorry for the lag

PR #16, opened 2024-01-10
feat: include rule name as comment

fgvieira:
Since the jobs are named with a unique ID, it is not easy to figure out what rule is being run on a specific SLURM job.
This PR adds the rule name as a comment, so that it can be displayed by `squeue` and `sacct`.


fgvieira:
How can I specify a rule-specific comment through `slum_extra`? Do you think something like `--default-resources slurm_extra="--comment {job.name}"` work? Or I'd have to specify a `slurm_extra` on each rule?

cmeesters:
You cannot, and as pointed out in #17 I concede.

cmeesters (approved):
good idea

PR #26, opened 2024-01-29
fix: Use sacct syntax that is compatible with slurm < 20.11.0

### LINKED ISSUE #25: Plugin is not compatible with older SLURM versions
Looks like the plugins is not compatible with old(ish) SLURM versions after #9 was merged.

My cluster still runs on version `slurm 20.02.7` and when trying using the plugin I get the following error message:

```
The job status query failed with command: sacct -X --parsable2 --noheader --format=JobIdRaw,State --starttime now-2days --endtime now --name 1a8c4af2-e653-401a-ae36-6ca49f05ba3d
Error message: Invalid time specification (pos=3): now-2days
```

From a quick investigation, it looks like the `now[{+|-}count[seconds(default)|minutes|hours|days|weeks]]` syntax was introduced with [version 20.11.0](https://slurm.schedmd.com/archive/slurm-20.11.0/sacct.html).

Would it make sense to more clearly state the minimal SLURM version compatible with the plugin?
Is there any workaround that I can try?

Thanks!

dnerini:
Closes #25 

Tested with slurm 20.02.07

Not a great contribution in terms of code readability, but it'd ensure that the plugin works with older slurm versions.

### DIFF
--- a/snakemake_executor_plugin_slurm/__init__.py
+++ b/snakemake_executor_plugin_slurm/__init__.py
@@ -206,7 +206,7 @@ class Executor(RemoteExecutor):
                 (status_of_jobs, sacct_query_duration) = await self.job_stati(
                     # -X: only show main job, no substeps
                     f"sacct -X --parsable2 --noheader --format=JobIdRaw,State "
-                    f"--starttime {datetime.utcnow() - timedelta(days=2):%Y-%m-%dT%H} "
+                    f"--starttime {datetime.utcnow() - timedelta(days=2):%Y-%m-%dT%H:00} "
                     f"--endtime now --name {self.run_uuid}"
                 )
                 if status_of_jobs is None and sacct_query_duration is None:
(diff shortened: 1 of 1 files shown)


fgvieira:
slurm `2.11.0` is more than 3 years old. Wouldn't it be better to update it?

dnerini:
I agreed, it would be great, but unfortunately that is not an option in my case. 
I'm happy to keep using my fork. I just thought I would submit the PR in case someone else is facing the same issue.

cmeesters:
Yep, and I would support this PR rather than lose the idea. I would also hate to require from you @dnerini to update your fork, whenever a relevant feature or fix is merged upstream ... and I understand that some admins are very reluctant to update their slurm. Although they really should due to security fixes.

Please run black on your code and add the comment.

dnerini:
thanks @cmeesters for the feedback. I reformatted the code, added a comment, and changed to local time using `datetime.now()` (although it shouldn't make a difference for the purpose introduced in #9).

cmeesters (approved):
finally, after retriggering the workflow ...

PR #27, opened 2024-01-29
feat: support --qos as a resource

kdm9:
A cluster I am using requires setting `--qos` to enable jobs with longer runtimes. it would be convenient to be directly able to use `slurm_qos` as a resource, like one can do for partition & account, so I've implemented that. Also fixed some typos in the docs.

best
K

### DIFF
--- a/docs/further.md
+++ b/docs/further.md
@@ -25,7 +25,7 @@ Usually, it is advisable to persist such settings via a
 [configuration profile](https://snakemake.readthedocs.io/en/latest/executing/cli.html#profiles), which
 can be provided system-wide, per user, and in addition per workflow.
 
-## Ordinary ]MP jobs
+## Ordinary SMP jobs
 
 Most jobs will be carried out by programs which are either single core
 scripts or threaded programs, hence SMP ([shared memory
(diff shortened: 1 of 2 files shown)


fgvieira:
Can't you use `slurm_extra` to pass the qos?

cmeesters:
Essentially, this is all covered by `slurm_extra`. However, the documentation needs an update in that regard (and I do not seem to get to it all too soon) and there has been a bug.

Please update to snakemake v8.4.3 or higher, that should fix working with `slurm_extra`. Yet, there is another bug, see PR #2627 and the discussion to #2555.

Note, that the idea behind this executor is to provide all relevant submission flags, but to catch cluster relevant flags, like qos (where some admins hide their accelerators, instead of gres) or flags introduced by spank plugins or all the nifty detail flags, sbatch provides with `slurm_extra`. If we were to include all flags explicetly, we would have to include a hundred ifs in the code.

kdm9:
How about something like:
```
for resource, value in job.resources.items:
    if not resource.startswith("slurm_"):
        continue
    slurmarg = resource.replace("slurm_", "")
    args += f" --{slurmarg}={value}"
```
In other words, to submit with `sbatch --arg=value`, provide "slurm_arg=value" as a resource.

That should handle partition as it happens now (perhaps updating `job.resources["slurm_partition"]` using the existing logic), and would allow users direct control over the submission command.

cmeesters:
I am closing this pull request: The current release is working with `slurm_extra` as intended, offering to set qos.

PR #35, opened 2024-02-13
feat: add wildcards to output and comment

Ulthran:
Add rule wildcards to slurm output logs and job comments. Not sure if this will need extra checks for when `job` is a group instead of a rule.

### DIFF
--- a/snakemake_executor_plugin_slurm/__init__.py
+++ b/snakemake_executor_plugin_slurm/__init__.py
@@ -66,7 +66,7 @@ class Executor(RemoteExecutor):
         # snakemake_interface_executor_plugins.executors.base.SubmittedJobInfo.
 
         log_folder = f"group_{job.name}" if job.is_group() else f"rule_{job.name}"
-        
+
         try:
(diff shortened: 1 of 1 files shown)


cmeesters (approved):
Thank you for this contribution. Did not find the time for it, yet. You simply did it!

cmeesters (commented):
AttributeError: 'GroupJob' object has no attribute 'wildcards'

cmeesters:
I'm afraid, that's not possible this way. Also: is adding arbitrary wildcards really desirable? Or just the rule name?

Ulthran:
Can we only include the wildcards if it is a `Job` type and pass an empty string if it's a `GroupJob`? Tried implementing that in the most recent commit.

In our case, we would always want to have all wildcards included in the output names. Is your concern that there might be cases where there are too many wildcards for this to be reasonable?

cmeesters (approved):
passes, works as desired

jaicher:
This breaks pipelines where wildcards can include the directory separator "/".

It should be straightforward to fix by sanitizing the wildcard string used in the logfile name to remove/replace the extra directory separators.

PR #42, opened 2024-02-29
docs: added mini paragraph about Conda and Env Modules

cmeesters:
Just a mini change, writing hints about software usage.

### DIFF
--- a/docs/further.md
+++ b/docs/further.md
@@ -182,7 +182,7 @@ While Snakemake mainly relies on Conda for reproducible execution, many clusters
 
 Note, `snakemake --sdm conda ...` works as intended.
 
-To ensure the working of this plugin, install it in your base environment for the desired workflow.
+To ensure that this plugin is working, install it in your base environment for the desired workflow.
 
 
 ### Using Cluster Environment:  Modules
--- a/docs/further.md
+++ b/docs/further.md
@@ -144,7 +144,7 @@ Again, rather use a [profile](https://snakemake.readthedocs.io/en/latest/executi
 
 While Snakemake mainly relies on Conda for reproducible execution, many clusters impose file number limits in their "HOME" directory. You can resort to[micromamba](https://mamba.readthedocs.io/en/latest/user_guide/micromamba.html) in place of [mamba](https://mamba.readthedocs.io/en/latest/), which is recommended by Snakemake: Micromamba does not save all the package files, hence users do not have to clean up manually. In this case, you need to install your environment for a workflow manually. Alternatively, run `mamba clean -a` occasionally for your environments.
 
-Note, `snakemake --use-conda ...` works as intended.
+Note, `snakemake --sdm conda ...` works as intended.
 
 To ensure the working of this plugin, install it in your base environment for the desired workflow.
 
--- a/docs/further.md
+++ b/docs/further.md
@@ -173,7 +173,6 @@ rule myrule:
 
 Again, rather use a [profile](https://snakemake.readthedocs.io/en/latest/executing/cli.html#profiles) to specify such resources.
 
-<<<<<<< HEAD
 ## Software Recommendations
 
 ### Conda, Mamba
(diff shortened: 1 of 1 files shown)


docs/further.md
```diff
@@ -136,4 +136,27 @@ rule myrule:
...
 
-Again, rather use a [profile](https://snakemake.readthedocs.io/en/latest/executing/cli.html#profiles) to specify such resources.
\ No newline at end of file
+Again, rather use a [profile](https://snakemake.readthedocs.io/en/latest/executing/cli.html#profiles) to specify such resources.
+
+## Software Recommendations
+
+### Conda, Mamba, Micromamba
+
+While Snakemake mainly relies on Conda for reproducible execution, many clusters impose file number limits in their "HOME" directory. Using [micromamba](https://mamba.readthedocs.io/en/latest/user_guide/micromamba.html) in place of [mamba](https://mamba.readthedocs.io/en/latest/), which is recommended by Snakemake, might be a remedy: Micromamba does not save all the package files, hence users do not have to clean up manually.
```
johanneskoester, line 145: Yes, but we have to first add micromamba support to Snakemake. There is a PR, but it needs to be adapted to Snakemake 8. I will try to finalize that ASAP. The most sustainable approach will be to first create an interface for deployment plugins and then move the PR code into a new plugin.

johanneskoester: I have been in contact with them. Pixi is not really meant for usage with envs like micromamba/mamba/conda is. It is rather like a generalization of poetry. With one env per "project", and users not really interacting directly with envs. However, pixi builds on rattler, and I think it makes most sense to just use rattler-py (a python api to rattler) in Snakemake directly. I will soon work on the software-deployment plugin interface for snakemake, then move the conda support into a corresponding plugin and offer a rattler option there.

cmeesters: > Pixi is not really meant for usage with envs like ...

So I noticed. Still need to wrap my head around this and how the differences can be introduced in courses.

cmeesters: The last commit is an attempt to clarify.

docs/further.md
```diff
@@ -136,4 +136,32 @@ rule myrule:
...
\ No newline at end of file
+Again, rather use a [profile](https://snakemake.readthedocs.io/en/latest/executing/cli.html#profiles) to specify such resources.
+
+## Software Recommendations
+
+### Conda, Mamba, Micromamba
+
+While Snakemake mainly relies on Conda for reproducible execution, many clusters impose file number limits in their "HOME" directory. You can resort to[micromamba](https://mamba.readthedocs.io/en/latest/user_guide/micromamba.html) in place of [mamba](https://mamba.readthedocs.io/en/latest/), which is recommended by Snakemake: Micromamba does not save all the package files, hence users do not have to clean up manually. In this case, you need to install your environment for a workflow manually. Alternatively, run `mamba clean -a` occasionally for your environments.
+
+Note, `snakemake --use-conda ...` works as intended.
```
johanneskoester, line 147:
```suggestion
Note, `snakemake --sdm conda ...` works as intended.
```

docs/further.md
```diff
@@ -136,4 +136,32 @@ rule myrule:
...
+   ...
+   envmodules:
+       "bio/VinaLC"
+```
+
+This will, internally, trigger a `module load bio`/VinaLC` immediately prior to execution. 
+
+Note, that 
+- environment modules are best specified in a configuration file.
+- `--use-envmodules` can be combined with `--use-conda` and `--use-singularity`, which will then be only used as a fallback for rules not defining environment modules.
```
johanneskoester, line 167:
```suggestion
- Using envmodules can be combined with conda and apptainer (`--sdm envmodules conda apptainer`), which will then be only used as a fallback for rules not defining environment modules.
```

johanneskoester (approved):

cmeesters:
@johanneskoester ready for merge? Note, that this PR does not reflect anything in addition to software handling. I/we need to update the docs regarding other aspects, soon, anyway.

cmeesters:
@johanneskoester hopefully after all those changes from upstream, now ok. git scrambled the text considerably in between. Don't know why.

Issue #47, opened 2024-03-04
Directory separator in wildcards not handled in updated logfile names


jaicher:
#35 adds wildcards to logfile name. As mentioned in https://github.com/snakemake/snakemake-executor-plugin-slurm/pull/35#issuecomment-1977523028, if the wildcards include the dimension separator "/", the parent directory of the resulting logfile cannot be created prior to submitting to slurm/is done incorrectly, so these jobs immediately fail.

cmeesters:
Thanks for being patient: I was on a conference.

Can you please provide an example? I'm trying to figure out, whether your issue stems from the character itself, or the path separator (os.path.sep` on Linux). Eventually, the code should be as robust as possible (even though SLURM will not run under windows).

jaicher:
Hi @cmeesters thanks for being patient as well; I was out traveling. The issue comes from the path separator, not the character itself. When the path separator is present in `wildcard_str`, it makes `os.path.dirname(slurm_logfile)` include `%j`, so `os.makedirs(...)` cannot make the appropriate directory for the slurm logfile. I pushed a modification replacing "/" with `os.path.sep` to my PR. Thank you for recognizing the OS-specific logic and helping me improve the PR!

cmeesters:
Ought to be fixed with merger of PR#67, be sure to have version >= 0.4.4 .

PR #48, opened 2024-03-04
fix: replace os.path.sep in wildcard strings with "" in logfile name

jaicher:
This replaces "/" with "." in logfile name for wildcards to make sure we can create parent directory for output log file, fixing #47. I chose "." as it is different than the wildcard separator "_", and having the wildcard name be faithful isn't super critical because the name is unique from the job id anyway.

### DIFF
--- a/snakemake_executor_plugin_slurm/__init__.py
+++ b/snakemake_executor_plugin_slurm/__init__.py
@@ -69,6 +69,7 @@ class Executor(RemoteExecutor):
 
         try:
             wildcard_str = f"_{'_'.join(job.wildcards)}" if job.wildcards else ""
+            wildcard_str = wildcard_str.replace("/", ".")
         except AttributeError:
             wildcard_str = ""
 
--- a/snakemake_executor_plugin_slurm/__init__.py
+++ b/snakemake_executor_plugin_slurm/__init__.py
@@ -69,7 +69,7 @@ class Executor(RemoteExecutor):
 
         try:
             wildcard_str = f".{'_'.join(job.wildcards)}" if job.wildcards else ""
-            wildcard_str = wildcard_str.replace("/", ".")
+            wildcard_str = wildcard_str.replace(os.path.sep, "")
         except AttributeError:
             wildcard_str = ""
 
--- a/snakemake_executor_plugin_slurm/__init__.py
+++ b/snakemake_executor_plugin_slurm/__init__.py
@@ -68,7 +68,7 @@ class Executor(RemoteExecutor):
         log_folder = f"group_{job.name}" if job.is_group() else f"rule_{job.name}"
 
         try:
-            wildcard_str = f"_{'_'.join(job.wildcards)}" if job.wildcards else ""
+            wildcard_str = f".{'_'.join(job.wildcards)}" if job.wildcards else ""
             wildcard_str = wildcard_str.replace("/", ".")
         except AttributeError:
             wildcard_str = ""


snakemake_executor_plugin_slurm/__init__.py
```diff
@@ -69,6 +69,7 @@ def run_job(self, job: JobExecutorInterface):
 
         try:
             wildcard_str = f"_{'_'.join(job.wildcards)}" if job.wildcards else ""
```
fgvieira, line 71:
```suggestion
            wildcard_str = f".{'_'.join(job.wildcards)}" if job.wildcards else ""
```

snakemake_executor_plugin_slurm/__init__.py
```diff
@@ -69,6 +69,7 @@ def run_job(self, job: JobExecutorInterface):
...
+            wildcard_str = wildcard_str.replace("/", ".")
```
fgvieira, line 72: Why not remove it?
```suggestion
            wildcard_str = wildcard_str.replace("/", "")
```

Also, should we try to deal with other reserved characters?

cmeesters: I'm not sure about the purpose. Is it about removing the '/'? Or `os.path.sep`? In that case, we should be agnostic.

Also, I think we should get rid of the try/except clause, for a simple test will suffice. I will ask about this in the issue and act accordingly.

fgvieira: In my case the problem was that, since the `wildcards` contained a '/', `slurm` failed (I guess since the folder did not exist).
Not sure but, if there are other reserved characters in the `wildcards`, it might be also good to remove them.

johanneskoester: This is weird, because we do a makedirs. So the number of subdirs here should be no problem. Maybe rather a latency issue?

fgvieira: I think the problem is that, when creating that folder, the `%s` is not replaced yet. It is `slurm` that replaces it.:
```
slurm_logfile = ".snakemake/slurm_logs/rule_test/%j_wc1_wc21/wc22_wc3.log"
```
So `snakemake` will create the folder `.snakemake/slurm_logs/rule_test/%j_wc1_wc21` and slurm will try to write on `.snakemake/slurm_logs/rule_test/123456789_wc1_wc21/wc22_wc3.log`, no?

jaicher: Exactly as @fgvieira stated -- the %j wildcard with which Slurm injects the jobid is not known when we make directories, so it cannot be used as in a parent directory for the final logfile.

In terms of reserved characters, I think we only need to worry about os.path.sep. Since wildcards themselves are extracted from other paths, I would expect them not to contain any other problematic wildcards.

cmeesters: @fgvieira ah, thanks for the elaboration. I think we need a pre-configured issue report here, too. Those minimal examples really help a lot!

fgvieira: @cmeesters any suggestions?

johanneskoester: I think the best approach is to extend the testcases that we already have, so that all executors benefit. Besides, I think I have fixed the issue in PR https://github.com/snakemake/snakemake-executor-plugin-slurm/pull/67 in a different way. The %j is now in the filename, so that we can safely create the dir before.

johanneskoester: And https://github.com/snakemake/snakemake-executor-plugin-slurm/pull/69 for a first try with that updated testcase.

johanneskoester: Seems to work well. Once https://github.com/snakemake/snakemake/pull/2810 is merged, this will be tested automatically.

johanneskoester: Am I right that this PR is then not necessary anymore?

cmeesters: Honestly, I was checking on this PR. Yes, we can close in in favour of #67

Issue #53, opened 2024-03-20
Missing support for --clusters option


meliamne:
Hi,

Our hpc runs 2 slurm clusters, this requires us to provide the -M or --cluster option to most slurm related commands like sbatch, squeue and scancel. This option is unsupported in snakemake 8+ as far as i can tell. In older snakemake versions support existed through profiles.

I can get submission to work by specifying --cluster in slurm_extra. However, snakemake is unable to detect when a job finishes and start the next set of rules.

slurm: 23.02.7
snakemake: 8.9.0
snakemake-executor-plugin-slurm: 0.4.2

cmeesters:
Actually, the idea is: provide the most-common flags, wrap everything else in `slurm_extra` (documentation is forthcoming).

However, I will discuss adding `cluster` and `gres` now.

meliamne:
Thank you very much for looking into this.
I think the current approach generally makes a lot of sense. The main issue in our HPC's case is that, as far as i can tell, the `--cluster` option from `slurm_extra` is only applied when submitting jobs. However, it is not applied when checking the status of a job with `sacct` or canceling a job with `scancel`. In our environment running these commands without the `--cluster` option specified returns nothing. Thus after starting the first round of jobs, snakemake gets stuck trying to check the status of the jobs.

cmeesters:
Please check, whether PR #56 fixes your issue: I do not have access to a multicluster setup. That might change in the future, but for now, I cannot test this feature.

And you are right: Checking the job status and cancelling jobs, has to be done be the executor plugin. If the information is only available at submit time, this would not be sufficient.

cmeesters:
meanwhile merged. Sorry, it took so long. Stuff happened. Please test version 0.9 once released on bioconda or pypi.

PR #56, opened 2024-04-02
feat: multicluster

cmeesters:
Putative fix for issue #53

### DIFF
--- a/snakemake_executor_plugin_slurm/__init__.py
+++ b/snakemake_executor_plugin_slurm/__init__.py
@@ -148,6 +148,7 @@ class Executor(RemoteExecutor):
         call += f" --cpus-per-task={get_cpus_per_task(job)}"
 
         if job.resources.get("slurm_extra"):
+            self.check_slurm_extra(job)
             call += f" {job.resources.slurm_extra}"
 
         exec_job = self.format_job_exec(job)
@@ -502,3 +503,15 @@ class Executor(RemoteExecutor):
             "'slurm_partition=<your default partition>'."
         )
         return ""
+
+    def check_slurm_extra(self, job):
+        jobname = re.compile(r"--job-name[=?|\s+]|-J\s?")
+        if re.search(jobname, job.resources.slurm_extra):
+            raise WorkflowError(
+                "The '--job-name' option is not allowed in the 'slurm_extra' "
+                "parameter. The job name is set by snakemake and must not be "
+                "overwritten. It is internally used to check the stati of all "
+                "submitted jobs by this workflow."
+                "Please consult the documentation if you are unsure how to "
+                "query the status of your jobs."
+            )
--- a/snakemake_executor_plugin_slurm/__init__.py
+++ b/snakemake_executor_plugin_slurm/__init__.py
@@ -71,7 +71,7 @@ class Executor(RemoteExecutor):
         group_or_rule = f"group_{job.name}" if job.is_group() else f"rule_{job.name}"
 
         try:
-            wildcard_str = f"_{'_'.join(job.wildcards)}" if job.wildcards else ""
+            wildcard_str = "_".join(job.wildcards) if job.wildcards else ""
         except AttributeError:
             wildcard_str = ""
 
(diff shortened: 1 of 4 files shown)


snakemake_executor_plugin_slurm/__init__.py
```diff
@@ -88,6 +90,10 @@ def run_job(self, job: JobExecutorInterface):
         call += self.get_account_arg(job)
         call += self.get_partition_arg(job)
 
+        if job.resources.get("clusters"):
+            call += f" --clusters {job.resources.clusters}"
+            self._clusters = job.resources.clusters
```
johanneskoester, line 95: What happens if there are multiple jobs that select different clusters? To me this seems like that the check_active_jobs command will always take the clusters setting from the latest submitted job, which can't be correct in the general case.

johanneskoester (changes requested):

snakemake_executor_plugin_slurm/__init__.py
cmeesters: Right. While SLURM itself supports submitting to multiple clusters, with the current scheme we would need to record a list or strictly speaking, a set of clusters, which we constantly query: i.e. if cluster 'A' is selected for job 'a' and cluster 'B' for job 'b', there status query has to go to 'A,B'.

Would that constitute an acceptable solution?

meliamne:
when specifying the cluster in sbatch the output changes to `Submitted batch job 61317320 on cluster wice`. This causes line 162 to mistakenly pick up the name of the cluster instead of the jobid.

https://github.com/snakemake/snakemake-executor-plugin-slurm/blob/7103c7eb1d48ecae568e1b9a92b79c9cb6b226d5/snakemake_executor_plugin_slurm/__init__.py#L162

snakemake_executor_plugin_slurm/__init__.py
meliamne: sacct also accepts `--clusters all` this might be a convenient solution, maybe slightly less precise.

cmeesters:
@meliamne Thank you, your suggestion went into my last commit.

snakemake_executor_plugin_slurm/__init__.py
cmeesters: Thank you, easy enough! This went into 67cfc6c286564b710402d64ad679acbd378758a7 .

However, we do not have a test for it — yet.

cmeesters:
@johanneskoester does the current state find your approval? 

I dropped
- `self.__clusters` as an attribute, instead
- `sacct` and `scancel` use `--clusters=all`, as suggested.
- in addition, please take a look at the regex use in line 159 .

cmeesters:
@johanneskoester we do not have a test, yet (due to faulty multicluster setup in the test-engine). Should we merge regardless, considering that the test will take considerable time and effort to implement?

cmeesters:
@johanneskoester not sure, isn't the requested change already in there, resp. the issue solved?

cmeesters:
@johanneskoester please have a look, after all this rebasing and fixing, I am a little puzzled. It's embarrassing, but it took too long.

cmeesters:
We (@johanneskoester ) and me discussed that we do not have a solution for a multicluster test at hand for the time being. Hence, this PR is merged without a dedicated test.

I am sorry, I could merge this sooner. Too many things kept coming up, then holidays ....

nigiord:
Thanks for this merge! I think there is a mistake in the documentation though:

https://github.com/snakemake/snakemake-executor-plugin-slurm/blob/03d474f74f703c4e4530bbeda8f16dcf43764263/docs/further.md?plain=1#L122

The snakemake resources seems to be `clusters` (with an s), and not `cluster`:

https://github.com/snakemake/snakemake-executor-plugin-slurm/blob/03d474f74f703c4e4530bbeda8f16dcf43764263/snakemake_executor_plugin_slurm/__init__.py#L140

I also made a comment regarding the way jobid is now parsed: https://github.com/snakemake/snakemake-executor-plugin-slurm/issues/121#issuecomment-2270905243

Cheers,

cmeesters:
Ah, thanks. Will fix this asap (= not today).

PR #66, opened 2024-04-11
refactor: get cpus_per_task from jobstep executor

johanneskoester:


cmeesters:
will re-run after the merge in the jobstepexecutor is settled for a while

cmeesters:
@johanneskoester na, whilst the import error is fixed after the jobstep executor release, we seemingly run into other issues.

johanneskoester:
Looking into that. Seems like sbatch requests less cpus than the group needs (1 instead of 2).

PR #67, opened 2024-04-12
fix: always create logdir before sbatch

johanneskoester:


cmeesters:
Interesting solution!

Issue #75, opened 2024-04-19
Setting memory twice when submitting with slurm executor


kwells4:
## Versions
`snakemake` version 8.10.7
`snakemake-executor-plugin-slurm` version 0.4.4
`snakemake-executor-plugin-slurm-jobstep` version 0.2.1

## The problem
I am working on getting `snakemake` version 8 to work on my slurm server and keep getting the following error:

```
srun: fatal: SLURM_MEM_PER_CPU, SLURM_MEM_PER_GPU, and SLURM_MEM_PER_NODE are mutually exclusive.
```

I can see that two resource arguments are being passed when looking at the rule description:

```
[Fri Apr 19 13:42:50 2024]
rule fastqc:
    input: /pl/active/Anschutz_BDC/analysis/wells/analysis/wells/testing_snakemake8_slurm/raw_data/control_1_R1.fastq.gz, /pl/active/Anschutz_BDC/analysis/wells/analysis/wells/testing_snakemake8_slurm/raw_data/co
ntrol_1_R2.fastq.gz
    output: /pl/active/Anschutz_BDC/analysis/wells/analysis/wells/testing_snakemake8_slurm/results/fastqc_pre_trim/fastqc_control_1_summary_untrimmed.txt
    jobid: 0
    reason: Forced execution
    wildcards: results=/pl/active/Anschutz_BDC/analysis/wells/analysis/wells/testing_snakemake8_slurm/results, sample=control_1
    resources: mem_mb=15259, mem_mib=7630, disk_mb=43311, disk_mib=41305, tmpdir=<TBD>, slurm_partition=acompile, slurm_account=amc-general, slurm_extra=--output=/pl/active/Anschutz_BDC/analysis/wells/analysis/we
lls/testing_snakemake8_slurm/results/logs/fastqc_pre_trim/fastqc_control_1_untrimmed.out --error=/pl/active/Anschutz_BDC/analysis/wells/analysis/wells/testing_snakemake8_slurm/results/logs/fastqc_pre_trim/fastqc_
control_1_untrimmed.err --qos=compile, runtime=60, mem=16GB
```

However, I don't know where the `mem_mb` is being passed. 

### Profile

```yaml
executor: slurm

default-resources:
    slurm_partition: "acompile"
    slurm_account:   "amc-general"

set-resources:
    fastqc:
        runtime: 60 # 1 hour
        mem: "16GB"
    fastqc_summary:
        runtime: 10
        mem: "4GB"
[...]

Issue #22, opened 2024-01-19
`SLURM_*` environment variables can interfere with execution of MPI workloads


ocaisa:
I have seen that if launching a workflow from within the context of a SLURM job (and therefore with plenty of `SLURM_*` environment variables set), this can interfere with the correct execution of an MPI application. In my case, I only observed this when executing across multiple nodes from within JupyterHub (where the Jupyter environments run on a compute node). I found that unsetting all the SLURM variables before running snakemake allows the workflow to complete correctly. To do this, you can use:
```
# Unset all environment variables that start with SLURM_
for var in $(printenv | grep '^SLURM_' | awk -F= '{print $1}'); do unset $var; done
```
before running your `snakemake` commands.

I'm not sure if this is something that could/should be fixed here or whether it should be resolved within the cluster environment, but I put it here in case someone else comes across the issue.

cmeesters:
I'm not sure, whether this is an issue of snakemake. Consider, that the default SLURM behaviour is to inherit the environment of your local set up (e.g. on login or head nodes) for your jobs. This will likely be the case for your Jupyter-Node, too. 

In other words: It's a quirk of your environment.

ocaisa:
Yes, not expecting a fix here (so I will close this., I was mainly reporting it in case someone else comes across the same issue. I can indeed start the workflow from a login node, but that would be outside the JupyterLab context.

PR #78, opened 2024-05-02
feat: warning if run in job

cmeesters:
Already, two issues (#75 and #22) seem to result from running Snakemake in a SLURM job context and using this executor plugin. This PR introduces detecting if triggered within a SLURM job and issuing a warning accordingly.

In principle, the plugin may work in job context. Submitting jobs from jobs has always been a highlight of SLURM. However, settings may lead to unintended behaviour (and would do so without Snakemake, presumably). Hence, we can only warn from the executor.

### DIFF
--- a/docs/further.md
+++ b/docs/further.md
@@ -2,7 +2,7 @@
 
 ## The general Idea
 
-To use this plugin, log in to your cluster's head node (sometimes called the "login" node), activate your environment as usual and start Snakemake. Snakemake will then submit your jobs as cluster jobs.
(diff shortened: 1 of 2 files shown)


cmeesters:
@johanneskoester I think this is a rather trivial change - however, I do not know how to extend the test suite for it. Pointers? Or can it be merged like that?

cmeesters:
@johanneskoester I consider this an urgent PR - there are always people you try something like [this](https://github.com/snakemake/snakemake/issues/2912).

cmeesters:
waited long enough

Issue #76, opened 2024-04-25
Make the `slurm_account` parameter optional, because it is in fact optional.


xapple:
Some SLURM installations don't require specifying an account when using `sbatch` or `salloc`.

Currently the value `-A (null)` is actually used in the `sbatch` call if you don't mention one:

    rule xxxxxxxxxxxx:
        output: xxxxxxxxxxxxxx
        jobid: 4
        reason: Missing output files: xxxxxxxxxxxxxx
        wildcards: dataset=xxxxxxxxxxxxxx, sample=xxxxxxxxxxxxxx
        resources: mem_mb=9193, mem_mib=8768, disk_mb=1000, disk_mib=954, tmpdir=<TBD>, slurm_partition=cpu, nodes=1, cpus_per_task=1, runtime=60

    No SLURM account given, trying to guess.
    Guessed SLURM account: (null)
    WorkflowError:
    SLURM job submission failed. The error message was /bin/sh: 1: Syntax error: "(" unexpected

And if you set it to some placeholder value such as "default" for instance, this happens:

    rule xxxxxxxxxxxx:
        output: xxxxxxxxxxxxxx
        jobid: 4
        reason: Missing output files: xxxxxxxxxxxxxx
        wildcards: dataset=xxxxxxxxxxxxxx, sample=xxxxxxxxxxxxxx
        resources: mem_mb=9193, mem_mib=8768, disk_mb=1000, disk_mib=954, tmpdir=<TBD>, slurm_partition=cpu, nodes=1, cpus_per_task=1, runtime=60

    WorkflowError:
    Unable to test the validity of the given or guessed SLURM account 'default' with sacctmgr: You are not running a supported accounting_storage plugin
    (accounting_storage/filetxt).
    Only 'accounting_storage/slurmdbd' and 'accounting_storage/mysql' are supported.

cmeesters:
Just taught a course on a cluster, with a default resp. None-account. Worked like a charm. 

So, I wonder, what is the output of `sacct -nu $USER -o Account%256` for you?

xapple:
Here is the output:

```
                                                                                                                                                                                                                                                          (null)
                                                                                                                                                                                                                                                          (null)
                                                                                                                                                                                                                                                          (null)
                                                                                                                                                                                                                                                          (null)
                                                                                                                                                                                                                                                          (null)
                                                                                                                                                                                                                                                          
```

This pattern repeats for 702 lines.

[...]

PR #81, opened 2024-05-03
fix: null byte account guess

cmeesters:
attempt to ensure that a '(null)' byte does not overthrow the account testing.


cmeesters:
The issuer reported in #76 that this fix appears to be working.

cmeesters:
waited too long

PR #86, opened 2024-05-06
fix: allowing for accounts containing whitespace

cmeesters:
As [reported](https://github.com/snakemake/snakemake/issues/2320), the plugin will not deal with account names containing whitespace. (accounts containing whitespace were not anticipated).

This PR is an attempt to fix this.

### DIFF
--- a/snakemake_executor_plugin_slurm/__init__.py
+++ b/snakemake_executor_plugin_slurm/__init__.py
@@ -369,7 +369,7 @@ class Executor(RemoteExecutor):
             # here, we check whether the given or guessed account is valid
             # if not, a WorkflowError is raised
             self.test_account(job.resources.slurm_account)
-            return f" -A {job.resources.slurm_account}"
+            return f" -A '{job.resources.slurm_account}'"
         else:
             if self._fallback_account_arg is None:
                 self.logger.warning("No SLURM account given, trying to guess.")
@@ -436,7 +436,7 @@ class Executor(RemoteExecutor):
                 f"'{account}' with sacctmgr: {e.stderr}"
             )
 
-        accounts = accounts.split()
+        accounts = [_.strip() for _ in accounts.split('\n') if len(_)]
 
         if account not in accounts:
             raise WorkflowError(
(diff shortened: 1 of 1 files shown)


snakemake_executor_plugin_slurm/__init__.py
```diff
@@ -440,7 +440,7 @@ def test_account(self, account):
                 f"'{account}' with sacctmgr: {e.stderr}"
             )
 
-        accounts = accounts.split()
+        accounts = [_.strip() for _ in accounts.split("\n") if _]
```
fgvieira, line 443: Not sure why but, in my HPC, the account appears repeated 5 times. Not sure if it might affect anything, but just in case maybe we can do:
```suggestion
        accounts = set([_.strip() for _ in accounts.split("\n") if _])
```

cmeesters: ugly, but yes! I will do so after a break and add a comment.

snakemake_executor_plugin_slurm/__init__.py
```diff
@@ -442,7 +442,10 @@ def test_account(self, account):
                 f"'{account}' with sacctmgr: {e.stderr}"
             )
 
-        accounts = [_.strip() for _ in accounts.split("\n") if _]
+        # The set() has been introduced during review. It eleminates
+        # duplicates. They are not harmful, but disturbing to read.
+        # the set eliminates duplicates - not harmful, but disturbing to read
```
fgvieira, line 447:
```suggestion
        # The set() has been introduced during review. It eleminates
        # duplicates. They are not harmful, but disturbing to read.
        # the set eliminates duplicates - not harmful, but disturbing to read
```
```suggestion
        # The set() has been introduced during review to eliminate
        # duplicates. They are not harmful, but disturbing to read.
```

snakemake_executor_plugin_slurm/__init__.py
cmeesters: seems to be working

cmeesters:
Thanks for stepping in and the excellent review!

snakemake_executor_plugin_slurm/__init__.py
fgvieira:
```suggestion
        # The set() has been introduced during review to eliminate
        # duplicates. They are not harmful, but disturbing to read.
```

cmeesters:
PS @fgvieira care to take a look again?

fgvieira (approved):

Issue #85, opened 2024-05-04
Wildcards missing from comments


mrvollger:
Hello, 

I was very excited to see in the [patch notes](https://github.com/snakemake/snakemake-executor-plugin-slurm/blob/main/CHANGELOG.md#features) the inclusion of wildcards in the slurm comment and log path.  

So I updated the plugin to get those features, and while I see wildcards in the log path (e.g. `.snakemake/slurm_logs/rule_fire/PS00474_HG02630_1_JAHAOQ010000001.1/18033650.log`), I still don't see them in the SLURM comment field, just the rule name (e.g. `fire`). Any ideas?

Thanks in advance,
Mitchell

cmeesters:
Hello Mitchell,

yes, indeed. We could do that. At first I was a bit hesitant - not knowing what the buffer size for SLURM strings is. And I bloody well do not want to break anybodies batch system. Yet, SLURM matured over time and as far as I can tell (please, anyone correct me, if I am wrong!), there is no such string limit for comments of job names. And my own tinkering with hello world scripts confirms this.

So, I will - as time permits - draft a PR.

Cheers
Christian

mrvollger:
Great! Made a small PR ^. Hope it helps.

mrvollger:
closed in #88.

cmeesters:
Ah, you are faster than me - got pretty busy.

Anyway, this issues does not go without a "Thank you!".

PR #88, opened 2024-05-06
feat: wildcards in comment string #85

mrvollger:
Small change to include wildcards in the comment strings (re #85). 

I tested it locally on my Slurm instance, and it works well. 

I also prefixed the rule name with `rule_` to match the formatting of the slurm log path. 

Cheers,
Mitchell

### DIFF
--- a/snakemake_executor_plugin_slurm/__init__.py
+++ b/snakemake_executor_plugin_slurm/__init__.py
@@ -89,7 +89,10 @@ class Executor(RemoteExecutor):
         # generic part of a submission string:
         # we use a run_uuid as the job-name, to allow `--name`-based
         # filtering in the job status checks (`sacct --name` and `squeue --name`)
-        comment_str = f"{job.name}_{wildcard_str}"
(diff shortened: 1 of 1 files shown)


snakemake_executor_plugin_slurm/__init__.py
```diff
@@ -89,9 +89,13 @@ def run_job(self, job: JobExecutorInterface):
         # generic part of a submission string:
         # we use a run_uuid as the job-name, to allow `--name`-based
         # filtering in the job status checks (`sacct --name` and `squeue --name`)
+        if wildcard_str == "":
+            comment_str = f"rule_{job.name}"
+        else:
+            comment_str = f"rule_{job.name}__wildcards_{wildcard_str}"
```
cmeesters, line 95: Other than the doubled underscore in this line, it looks fine.

mrvollger: Done

cmeesters (approved):
fine - let's try it

Issue #97, opened 2024-06-03
Batch job submission failed: Node count specification invalid


KB1RD:
Using the following Snakefile:
```py
rule ALL:
    resources:
        nodes=1
    run:
        shell('echo Hello World > test.txt')
```
And running: (where <act_name> is my account name)
```
snakemake -j 1 --executor slurm --default-resources slurm_account=<act_name>
```
I get:
```
Building DAG of jobs...
SLURM run ID: a624be26-2574-4b8b-9b59-516f07c62cb9
Using shell: /usr/bin/bash
Provided remote nodes: 1
Job stats:
job      count
-----  -------
ALL          1
total        1

Select jobs to execute...
Execute 1 jobs...

[Mon Jun  3 12:24:56 2024]
rule ALL:
    jobid: 0
    reason: Rules with neither input nor output files are always executed.
    resources: mem_mb=1000, mem_mib=954, disk_mb=1000, disk_mib=954, tmpdir=<TBD>, slurm_account=st-wangalex-1, nodes=1

No wall time information given. This might or might not work on your cluster. If not, specify the resource runtime in your rule or as a reasonable default via --default-resources.
WorkflowError:
SLURM job submission failed. The error message was sbatch: error: ---------------------------------------------------------------------- 
sbatch: error:  No number of nodes specified.  
sbatch: error: Please add  #SBATCH --nodes=NN  where NN is number of nodes for the job, in your submission script
sbatch: error: or in your interactive submission add --nodes=NN to your command.    
sbatch: error:  ---------------------------------------------------------------------
sbatch: error: Batch job submission failed: Node count specification invalid
```

I also replaced `sbatch` in my local bash path and used it to `echo` out the arguments that are passed to `sbatch`:
```
[...]

PR #98, opened 2024-06-04
fix: #97 preventing node confinment

cmeesters:
fixes the bug described in #97


fgvieira (approved):

fgvieira:
I have not tested the changes but it looks good. :smile:

Issue #117, opened 2024-07-22
The account or account/partition combination is not recognized!


Zhongzheng99:
I encountered a problem where I provided the correct partition and account name, but they seem to be unrecognized by the system. Here’s the command I used:
`snakemake --workflow-profile /home/u2204084007/RNA/cofig/profile --configfile cofig/config.yaml --profile profile`

/home/u2204084007/RNA/cofig/profile/config.yaml：
> default-resources:
>     slurm_partition: "c01"
>     slurm_account:   "u2204084007"

/home/u2204084007/.config/snakemake/profile/config.yaml:
> executor: slurm
> jobs: 100

The output indicates that the profiles are being used for setting default command line arguments:
```
Using profiles profile and workflow specific profile /home/u2204084007/RNA/cofig/profile for setting default command line arguments.
Building DAG of jobs...
```
However, there’s a warning that Snakemake is being run in a SLURM job context, which is not recommended:
```
You are running snakemake in a SLURM job context. This is not recommended, as it may lead to unexpected behavior. Please run Snakemake directly on the login node.
SLURM run ID: e4aa40d6-e26b-4558-a3f8-d614ef095b92
```
The system proceeds to list the jobs and selects 9 jobs to execute. One of the jobs, copy_input, is detailed with the specified resources, including the partition and account:
```
rule copy_input:
    input: rawdata/T21-1R_FRAS190282952-1a_1.clean.fq.gz
    output: fastq/raw/T21_1.L001.R1.fastq.gz
    jobid: 28
    reason: Missing output files: fastq/raw/T21_1.L001.R1.fastq.gz; Code has changed since last execution
    wildcards: sample=T21_1, lane=L001, pair=R1
    resources: mem_mb=5622, mem_mib=5362, disk_mb=5622, disk_mib=5362, tmpdir=<TBD>, slurm_partition=c01, slurm_account=u2204084007, io=1
```
Despite providing the correct partition (c01) and account (u2204084007), the job submission fails with a WorkflowError:
```
No wall time information given. This might or might not work on your cluster. If not, specify the resource runtime in your rule or as a reasonable default via --default-resources.
WorkflowError:
SLURM job submission failed. The error message was sbatch: error: invalid account or account/partition combination specified
sbatch: error: Batch job submission failed: Unspecified error
```
This error suggests that there is an issue with the account or partition combination, even though they appear to be correctly specified.

Zhongzheng99:
[...]

PR #118, opened 2024-07-22
Make sure sbatch receives --comments correctly

Zhongzheng99:
I’ve resolved the issue, and it turns out that the problem was not with the account and partition not being input correctly. The actual issue was with the `--comment` parameter in the sbatch command. In the original `__init__.py` file, line 133 was `f"--comment {comment_str}"`. It’s crucial to note that `comment_str` needs to be enclosed in quotes. If this is not done, it might cause sbatch to misinterpret `comment_str`, leading to the -A and -p parameters (which specify the account and partition) being incorrectly associated with the `--comment` parameter. This would result in the following WorkflowError:

> SLURM job submission failed. The error message was sbatch: error: invalid account or account/partition combination specified
> sbatch: error: Batch job submission failed: Unspecified error

To fix the issue, the line in the __init__.py file should be modified to ensure that comment_str is properly quoted,  like so:
```f"--comment=`{comment_str}`"```
Note: it’s important to use the` = `sign to properly assign the value to the `--comment` option in the sbatch command. 

By making this change, sbatch will correctly interpret the --comment parameter and subsequent -A and -p parameters, allowing the job to be submitted without errors.

The final command passed to sbatch would then be correctly formatted, like so:
```sbatch --comment=`Y{comment_str}` -A your_account -p your_partition ...```

### DIFF
--- a/snakemake_executor_plugin_slurm/__init__.py
+++ b/snakemake_executor_plugin_slurm/__init__.py
@@ -130,7 +130,7 @@ class Executor(RemoteExecutor):
             comment_str = f"rule_{job.name}_wildcards_{wildcard_str}"
         call = (
             f"sbatch --job-name {self.run_uuid} --output {slurm_logfile} --export=ALL "
-            f"--comment {comment_str}"
+            f"--comment `{comment_str}`"
         )
 
         call += self.get_account_arg(job)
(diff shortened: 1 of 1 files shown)


Zhongzheng99:
#117

Zhongzheng99:
Ah, this is a mistake because my account is in arrears.

PR #122, opened 2024-08-01
fix: #121 by more robustly parsing the output returned by sbatch

### LINKED ISSUE #121: Jobid is not properly parsed when cluster name is present
>   snakemake-executor-plugin-slurm=0.8.0
>   snakemake-executor-plugin-slurm-jobstep=0.2.1

Currently the executor-plugin parses the job id of the submitted jobs by taking the last word of the output from the sbatch command:

https://github.com/snakemake/snakemake-executor-plugin-slurm/blob/6a5ed46d6b4d692863ce981e6fff81915c0bd2fc/snakemake_executor_plugin_slurm/__init__.py#L194-L203

It fails on the cluster I use because for some reason the output looks like this:
> Submitted batch job 3969755 on cluster nautilus

Hence, the executor-plugin thinks that all my job ids are equal to `nautilus`. It raises multiple issues, including the fact that the scheduler cannot resume after jobs are completed.

I’m not sure why the message on my cluster is different from the one expected, I think it’s the default behaviour of slurm when a cluster name is defined (we have multiple clusters defined). My slurm version is:

`sinfo --version`
> slurm 22.05.7-Atos.1.1

In any case, the current parsing does not seem very robust as this output message could be altered in future versions of slurm. A temporary workaround on my side is to replace `slurm_jobid = out.split(" ")[-1]`by `slurm_jobid = out.split(" ")[-4] ` but that’s rather hideous. In recent versions of slurm, sbatch as the `--parsable` option that only returns jobid and clustername, semicolon separated:

https://slurm.schedmd.com/sbatch.html

> --parsable
>     Outputs only the job id number and the cluster name if present. The values are separated by a semicolon. Errors will still be displayed. 

Would it be possible to include this option directly in the sbatch call? And then to split by semicolon the result to get the job id (it should not failed, even if the cluster is not defined).

nigiord:
sbatch output was not properly parsed for job id when a cluster name is defined on the HPC. This fixes the issues by exploiting the `--parsable` option of sbatch.

### DIFF
--- a/snakemake_executor_plugin_slurm/__init__.py
+++ b/snakemake_executor_plugin_slurm/__init__.py
@@ -129,7 +129,7 @@ class Executor(RemoteExecutor):
         else:
             comment_str = f"rule_{job.name}_wildcards_{wildcard_str}"
         call = (
-            f"sbatch --job-name {self.run_uuid} --output {slurm_logfile} --export=ALL "
+            f"sbatch --parsable --job-name {self.run_uuid} --output {slurm_logfile} --export=ALL "
             f"--comment {comment_str}"
         )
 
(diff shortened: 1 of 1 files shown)


cmeesters:
rejected in favour of #56 , explained in #121

PR #125, opened 2024-08-06
fix: add --parsable to sbatch call for a more robust output parsing

nigiord:
Following discussion in https://github.com/snakemake/snakemake-executor-plugin-slurm/issues/121#issuecomment-2270905243 , this implementation seems more robust than match a word with only numbers (which could break in future versions of slurm, or if a cluster name is made of only numbers).


cmeesters (approved):

cmeesters:
Thank you for your contribution! I will merge this (after fixing the linter issue) - but a new release has to wait until some more contribs are merged. This contrib does not change behaviour, after all. It may wait a day or two.

nigiord:
Sorry for the linting issue, I fixed it. Of course the current implementation works fine so it can wait for the next release :) . Thank you for time!

cmeesters:
no need to apologize - thank you for your contribution once more!

(now, after the holidays, I am still pretty busy, but will work on improving the code once more)

PR #137, opened 2024-08-16
feat: in job stability

### LINKED ISSUE #113: Running snakemake in a SLURM job context
In our HPC environment I am not alowed to run anything on the login-nodes (kicked out, banned...). So I have to run snakemake ALWAYS with `sbatch` or using an interactive session via `srun`.

Therefore I always get the warning

```
You are running snakemake in a SLURM job context. This is not recommended, as it may lead to unexpected behavior.Please run Snakemake directly on the login node.
```

And the warning is correct. Because especially the number of threads are not correctly set.

Any idea how to run snakmake within a SLURM job properly? Slurm interplayed perfectly with snakemake 7 using `--cluster` `--clsuter-cancel` and so on. But now with snakemake 8 I am forced to run it with this plugin which does not work in our environment...

cmeesters:
might increase stability of in-job submissions, might fix #113 

## Summary by CodeRabbit

- **New Features**
  - Introduced a utility function to manage SLURM environment variables, enhancing job submission reliability.
  - Updated warning handling to clean up SLURM environments when specific conditions are met.

- **Bug Fixes**
  - Improved resource management by ensuring lingering SLURM job contexts are addressed.

### DIFF
--- a/snakemake_executor_plugin_slurm/__init__.py
+++ b/snakemake_executor_plugin_slurm/__init__.py
@@ -26,6 +26,8 @@ from snakemake_interface_executor_plugins.jobs import (
 from snakemake_interface_common.exceptions import WorkflowError
 from snakemake_executor_plugin_slurm_jobstep import get_cpus_per_task
 
+from utils import delete_slurm_environment
+
 
 @dataclass
 class ExecutorSettings(ExecutorSettingsBase):
@@ -85,10 +87,11 @@ class Executor(RemoteExecutor):
             if "SLURM_JOB_ID" in os.environ:
                 self.logger.warning(
                     "You are running snakemake in a SLURM job context. "
-                    "This is not recommended, as it may lead to unexpected behavior."
+                    "This is not recommended, as it may lead to unexpected behavior. "
                     "Please run Snakemake directly on the login node."
                 )
                 time.sleep(5)
+                delete_slurm_environment()
         done = True
 
     def additional_general_args(self):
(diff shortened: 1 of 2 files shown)


cmeesters:
@fgvieira would you have time to review this PR? Alas, we cannot test this within the CI (due to lack of resources). Seems to work fine for me and others.

fgvieira:
Why is it not recommended to run `snakemake` as a SLURM job? That is how I've been running it.

cmeesters:
Thing is: The executor exports the environment. This has to be done, because otherwise the base environment might not be present. Within a SLURM job the `SLURM_*` got exported, too. This caused SLURM in some cases to complain about `mem` and `mem_per_cpu` mismatches.

I am _not_ sure the current PR will fix all issues. I am even not sure, that 100 % stability can be achieved at all - apparently HPC admins made akward configuration a hobby ... But removing the scheduler env variables certainly is a step forward.

fgvieira:
So the idea would be to remove all `SLURM_*` env variables from the (new) launched job. Not the parent `snakemake` job, right?

cmeesters:
In fact, you cannot remove environment variables from a parent shell. Hence, python can only remove within the current environment on a node, not its jobscript. Within the Snakemake process on that node, you would not see the `SLURM_*` variables any more. Also, a daughter process of the same Python process would not be able to see them. A newly lauchned job (on the same node or a different one) would get a fresh population of  `SLURM_*`  env variables by SLURM itself.

If you use (host based) logins to that node, you too do not see these environment variables (because: new shell). Howver, if your run 

```shell
$ srun -A <account> -p <partition> --pty -t 10 bash -i
<node>:<path>$ python3
>>> import os
>>> print(os.environ["SLURM_JOB_ID"])
376803
>>> del os.environ["SLURM_JOB_ID"]
>>> print(os.environ["SLURM_JOB_ID"])
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "<frozen os>", line 714, in __getitem__
KeyError: 'SLURM_JOB_ID'
>>> 
<node>:<path>$ echo $SLURM_JOB_ID
376803
```
you will see, that you are able to tinker with env vars within your process, not in the parent shell.

fgvieira (approved):

PR #153, opened 2024-09-30
docs: requeue

cmeesters:
only documenting what is new in v0.11.0 (see #152 )

## Summary by CodeRabbit

- **New Features**
	- Introduced new command line flags for the SLURM executor plugin: `--slurm_init_seconds_before_status_checks` and `--slurm_requeue`.
- **Documentation**
	- Expanded guidance on job resubmission and retries, enhancing clarity on the use of `--retries` and `--slurm-requeue`.
	- Improved overall structure of the documentation for better understanding of job configuration and resource specifications.
	- Minor formatting corrections made for clarity.

### DIFF
--- a/docs/further.md
+++ b/docs/further.md
@@ -162,7 +162,8 @@ set-resources:
 
 ### Additional Command Line Flags
 
-This plugin defines additional command line flags. As always the can be used on the command line or in a profile.
+This plugin defines additional command line flags.
+As always, these can be set on the command line or in a profile.
 
 | Flag        | Meaning  |
 |-------------|----------|
--- a/docs/further.md
+++ b/docs/further.md
@@ -297,7 +297,7 @@ snakemake ... --rerun-incomplete
 snakemake ... --ri
 ```
 
-The "requeue" option allows jobs to be resubmitted automatically if they fail or are preempted. This might be the default on your cluster, already. You can check your cluster's requeue settings with 
+The "requeue" option allows jobs to be resubmitted automatically if they fail or are preempted. This is similar to Snakemake's `--retries`, except a SLURM job will not be considered failed and priority may be accumulated during pending. This might be the default on your cluster, already. You can check your cluster's requeue settings with 
 
 ```console
 scontrol show config | grep Requeue
--- a/docs/further.md
+++ b/docs/further.md
@@ -281,7 +281,7 @@ export SNAKEMAKE_PROFILE="$HOME/.config/snakemake"
 
 ==This is ongoing development. Eventually you will be able to annotate different file access patterns.==
 
-## <a name="retries"></a> Retries - Or Trying again when a Job failed
+## Retries - Or Trying again when a Job failed
 
 Some cluster jobs may fail. In this case Snakemake can be instructed to try another submit before the entire workflow fails, in this example up to 3 times:
 
--- a/docs/further.md
+++ b/docs/further.md
@@ -167,7 +167,7 @@ As always, these can be set on the command line or in a profile.
 
 | Flag        | Meaning  |
 |-------------|----------|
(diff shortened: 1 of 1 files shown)


cmeesters:
Would be nice if one of you would find the time to review. Thank you!

docs/further.md
```diff
@@ -160,6 +160,15 @@ set-resources:
         cpus_per_task: 40
 ```
 
+### Additional Command Line Flags
+
+This plugin defines additional command line flags. As always the can be used on the command line or in a profile.
```
dlaehnemann, line 165: Mini rephrasing of the second sentence. And I prefer having one sentence per line in Markdown files, for clearer git diffs down the line...
```suggestion
This plugin defines additional command line flags.
As always, these can be set on the command line or in a profile.
```

docs/further.md
```diff
@@ -160,6 +160,15 @@ set-resources:
...
+
+| Flag        | Meaning  |
+|-------------|----------|
+| `--slurm_init_seconds_before_status_checks`| will modify the default time (40 seconds) before the initial status check - usefull for development purposes|
```
dlaehnemann, line 169: Just expanding a bit on the reasoning behind the default and where shorter wait times might be desirable.
```suggestion
| `--slurm_init_seconds_before_status_checks`| modify time before initial job status check; the default of 40 seconds avoids load on querying slurm databases, but shorter wait times are for example useful during workflow development |
```

docs/further.md
```diff
@@ -271,7 +280,7 @@ export SNAKEMAKE_PROFILE="$HOME/.config/snakemake"
 
 ==This is ongoing development. Eventually you will be able to annotate different file access patterns.==
 
-## Retries - Or Trying again when a Job failed
+## <a name="retries"></a> Retries - Or Trying again when a Job failed
```
dlaehnemann, line 283: Are you sure, this manual html link is necessary? Usually, you should be able to link within a markdown document, directly referring to headings (with whitespace replaced by dashes `-`):
https://stackoverflow.com/questions/2822089/how-to-link-to-part-of-the-same-document-in-markdown

docs/further.md
```diff
@@ -282,7 +291,21 @@ snakemake --retries=3
...
+The "requeue" option allows jobs to be resubmitted automatically if they fail or are preempted. This might be the default on your cluster, already. You can check your cluster's requeue settings with 
+
+```console
+scontrol show config | grep Requeue
+```
+
+This requeue feature is integrated into the SLURM submission command, adding the --requeue parameter to allow requeuing after node failure or preemption using:
+
+```console
+snakemake --slurm-requeue ...
```
dlaehnemann, line 308: How is this different from snakemake's own `--retries` option? Or does this simply implement the `--retries` value for pre-empted jobs or does it simply move the retries to the level of the slurm scheduler? Maybe you can clarify this further?

dlaehnemann (changes requested):
Nice extra features, and great to have them documented. I have some questions and suggestions.

docs/further.md
cmeesters: thank you

docs/further.md
cmeesters: Right. This requires some SLURM background. I tried to explain in a single sentence in commit cb153d2e21f65def715c26f728e9cd98f9be4188 .

docs/further.md
cmeesters: let's see. The html is removed in ed0c0d0b02ba902504b5b0fb97bd6fb4c2e6c32f

dlaehnemann (approved):
Thanks for the updates!

cmeesters:
> Thanks for the updates!

Thank you for your prompt review and the lectoring!

Issue #123, opened 2024-08-03
Way to make logging go to same logs as when running locally? 


hermidalc:
Is there a way to configure the executor so job logs to the log locations specified in my rules? (i.e. just like when running my workflow locally outside of SLURM). So instead of everything going to `.snakemake/slurm_logs`

cmeesters:
no, but it has been asked sometimes. I'll add it on the to-do list.

hermidalc:
In hindsight it actually makes sense to me the way it is, SLURM specific snakemake logging, job info and e.g. restarts/retries etc go into `.snakemake/slurm_logs` and the regular rule log output of the successful execution of the rule goes into the rule log path. It's less clutter and you don't need to see the rule regular logging from failed execution attempts mixed in with SLURM specific logging if the job needed to restart for any reason.

cmeesters:
absolutely, that was the intend. However, the reason, why I leave this issue open is
- that the feature was requested multiple times
- that i want to feature optional deletion of SLURM outputs (they are mostly redundant and if a workflow suceeds, they can go away as the logging information by Snakemake contains all valuable info, plus the report, etc.)

cmeesters:
@hermidalc now enabled with the latest release. Check out the docs - and thank you for your patience. The time to this release was long ... live hit me.

Issue #94, opened 2024-05-23
option to set the log directory?


jbloom:
Is it possible to enable the log directory to be set to something other than a directory in `.snakemake`? For large workflows that submit many jobs, it would be nice to be able to write these to a scratch directory rather than putting tens of thousands of files in `.snakemake`.

bentyeh:
I'd like to second the above request and consolidate several previous requests across multiple issues for giving the user more flexibility in specifying how to log stdout and stderr from rules submitted as SLURM jobs:

1. Specify log directory (this issue, #94)
2. Specify which wildcards to use, not just concatenate all of them (#11, #34, #84)
3. Specify separate names and paths for stdout (to pass to `sbatch --output=...`) and stderr (to pass to `sbatch --error=...`) (#95)

More generally, I hope that this new executor framework can at least achieve feature parity with the previous version (e.g., v7) of Snakemake that supported a very flexible logging setup via `snakemake --cluster-config cluster.yaml ... <Snakefile>`

where `cluster.yaml` looked something like

```{yaml}
__default__:
    time: 01:00:00
    threads: 2
    mem: 10G
    error: "logs/{rule}.{wildcards.sample}.err"
    output: "logs/{rule}.{wildcards.sample}.out"
```

cmeesters:
@jbloom no, however, refactoring the logging feature (particularly, deleting log files older than n days) is on my TODO list. The log directory is a core feature - I will discuss this with internally. We will work on this, it might take a little while.

@bentyeh please do not mix issues. I ensure, that we take all feature requests seriously. ad 2) I can see your point, however, this too needs some discussion. After all, implementing it the way it is, was a feature request. And a change is not as straightforward as it seems: configurability must be maintained at three points: the command line, the Snakefiles and the profile. Adding a keep-those-wildcars-option for the CLI would be awkward, right?

Regarding your other critics: the workflow profile offers just that, albeit that I too would like to see something changed. It ought to be possible to specify a single file, not a directory.

jbloom:
@cmeesters, great, thank you so much for considering customizability of the log directory as a feature. That would be tremendously useful and I appreciate your time and consideration of this.

mbhall88:
Agreed, this would be very useful. 

A second thing I want to ask for is documenting where the logs are actually output to in the plugin docs (unless I've missed it).

cmeesters:
it's on my todo list

blaiseli:
[...]

PR #159, opened 2024-10-20
feat: custom log file behaviour

cmeesters:
This PR should allow to

- set a custom log directory for SLURM jobs
- enable auto-deletion of older SLURM log files after 10 days
- allow changing the that behaviour or disabling auto deletion all together
- auto delete log files of successful SLURM jobs (and ignore this default behaviour)

The idea behind this PR is to
- limit the number of SLURM log files to keep, as a workflow can easily produce thousands of log files. In case of a successful job, the information is redundant anyway, hence the proposed auto-deletion of successful jobs.
- users may select a custom directory for SLURM log files - this way different workflows can point to one prefix. Together with the auto-deletion of older log files, this futher limits the number of present log files.

It addresses issues #94 and #123.

## Release Notes

- **New Features**
  - Enhanced SLURM job log management with a configurable log directory.
  - Added options to control log file retention and automatic cleanup of old logs.
  - Introduced new command-line flags for specifying log management settings.

- **Documentation**
  - Updated documentation with new command-line flags for log management.
  - Improved guidance on managing job logs and their retention policies.

- **Improvements**
  - More flexible control over SLURM job log handling.
  - Better support for cleaning up empty directories after job completion.

### DIFF
--- a/docs/further.md
+++ b/docs/further.md
@@ -143,20 +143,6 @@ set-resources:
         cpus_per_task: 40
 ```
 
-#### Additional Command Line Flags
-
-This plugin defines additional command line flags.
-As always, these can be set on the command line or in a profile.
-
-| Flag        | Meaning  |
-|-------------|----------|
-| `--slurm-logdir` | can change the default directory for SLURM logs, which is `~/.snakemake/slurm_logs/`|
-| `--slurm-keep-successful-logs` | per default all SLURM logs of successful jobs will be deleted upon job completion, these are just too many files too keep. The flag will overwrite this behaviour.|
-| `--slurm-delete-logfiles-older-than` | all log files in `--slurm-logdir` will be deleted after some time (default is after 10 days). If set to <= 0 this deletion will be skipped. The flag allows to change the default.|
-| `--slurm-init-seconds-before-status-checks`| modify time before initial job status check; the default of 40 seconds avoids load on querying slurm databases, but shorter wait times are for example useful during workflow development |
-| `--slurm-requeue` | allows jobs to be resubmitted automatically if they fail or are preempted. See the [section "retries" for details](#retries)|
-
-
 #### Multicluster Support
 
 For reasons of scheduling multicluster support is provided by the `clusters` flag in resources sections. Note, that you have to write `clusters`, not `cluster`! 
--- a/docs/further.md
+++ b/docs/further.md
@@ -1,10 +1,8 @@
-# The Executor Plugin for HPC Clusters using the SLURM Batch System
-
-## The general Idea
+### The general Idea
 
 To use this plugin, log in to your cluster's head node (sometimes called the "login" node), activate your environment as usual, and start Snakemake. Snakemake will then submit your jobs as cluster jobs.
 
-## Specifying Account and Partition
+### Specifying Account and Partition
 
 Most SLURM clusters have two mandatory resource indicators for
 accounting and scheduling, the account and a
@@ -33,7 +31,7 @@ can be provided system-wide, per user, and in addition per workflow.
 
 The executor waits per default 40 seconds for its first check of the job status. Using `--slurm-init-seconds-before-status-checks=<time in seconds>` this behaviour can be altered.
 
-## Ordinary SMP jobs
+### Ordinary SMP jobs
 
 Most jobs will be carried out by programs that are either single-core
 scripts or threaded programs, hence SMP ([shared memory
@@ -61,7 +59,7 @@ rule a:
 ```
 instead of the `threads` parameter. Parameters in the `resources` section will take precedence.
 
-## MPI jobs
+### MPI jobs
 
 Snakemake\'s SLURM backend also supports MPI jobs, see
 `snakefiles-mpi`{.interpreted-text role="ref"} for details. When using
@@ -90,7 +88,7 @@ $ snakemake --set-resources calc_pi:mpi="mpiexec" ...
 
 To submit "ordinary" MPI jobs, submitting with `tasks` (the MPI ranks) is sufficient. Alternatively, on some clusters, it might be convenient to just configure `nodes`. Consider using a combination of `tasks` and `cpus_per_task` for hybrid applications (those that use ranks (multiprocessing) and threads). A detailed topology layout can be achieved using the `slurm_extra` parameter (see below) using further flags like `--distribution`.
 
-## Running Jobs locally
+### Running Jobs locally
 
 Not all Snakemake workflows are adapted for heterogeneous environments, particularly clusters. Users might want to avoid the submission of _all_ rules as cluster jobs. Non-cluster jobs should usually include _short_ jobs, e.g. internet downloads or plotting rules.
 
@@ -100,7 +98,7 @@ To label a rule as a non-cluster rule, use the `localrules` directive. Place it 
 localrules: <rule_a>, <rule_b>
 ```
 
-## Advanced Resource Specifications
+### Advanced Resource Specifications
 
 A workflow rule may support several
 [resource specifications](https://snakemake.readthedocs.io/en/latest/snakefiles/rules.html#resources).
@@ -160,7 +158,7 @@ set-resources:
         cpus_per_task: 40
 ```
 
-### Additional Command Line Flags
+#### Additional Command Line Flags
 
 This plugin defines additional command line flags.
 As always, these can be set on the command line or in a profile.
@@ -170,11 +168,11 @@ As always, these can be set on the command line or in a profile.
 | `--slurm_init_seconds_before_status_checks`| modify time before initial job status check; the default of 40 seconds avoids load on querying slurm databases, but shorter wait times are for example useful during workflow development |
 | `--slurm_requeue` | allows jobs to be resubmitted automatically if they fail or are preempted. See the [section "retries" for details](#retries)|
 
-## Multicluster Support
+#### Multicluster Support
 
 For reasons of scheduling multicluster support is provided by the `clusters` flag in resources sections. Note, that you have to write `clusters`, not `cluster`! 
 
-## Additional Custom Job Configuration
+#### Additional Custom Job Configuration
 
 SLURM installations can support custom plugins, which may add support
 for additional flags to `sbatch`. In addition, there are various batch options not directly supported via the resource definitions
@@ -191,9 +189,9 @@ rule myrule:
 
 Again, rather use a [profile](https://snakemake.readthedocs.io/en/latest/executing/cli.html#profiles) to specify such resources.
 
-## Software Recommendations
+### Software Recommendations
 
-### Conda, Mamba
+#### Conda, Mamba
 
 While Snakemake mainly relies on Conda for reproducible execution, many clusters impose file number limits in their "HOME" directory. In this case, run `mamba clean -a` occasionally for persisting environments.
 
@@ -202,7 +200,7 @@ Note, `snakemake --sdm conda ...` works as intended.
 To ensure that this plugin is working, install it in your base environment for the desired workflow.
 
 
-### Using Cluster Environment:  Modules
+#### Using Cluster Environment:  Modules
 
 HPC clusters provide so-called environment modules. Some clusters do not allow using Conda (and its derivatives). In this case, or when a particular software is not provided by a Conda channel, Snakemake can be instructed to use environment modules. The `--sdm env-modules` flag will trigger loading modules defined for a specific rule, e.g.:
 
@@ -220,7 +218,7 @@ Note, that
 - Using environment modules can be combined with conda and apptainer (`--sdm env-modules conda apptainer`), which will then be only used as a fallback for rules not defining environment modules.
 For running jobs, the `squeue` command:
 
-## Inquiring about Job Information and Adjusting the Rate Limiter
+### Inquiring about Job Information and Adjusting the Rate Limiter
 
 The executor plugin for SLURM uses unique job names to inquire about job status. It ensures inquiring about job status for the series of jobs of a workflow does not put too much strain on the batch system's database. Human readable information is stored in the comment of a particular job. It is a combination of the rule name and wildcards. You can ask for it with the `sacct` or `squeue` commands, e.g.:
 
@@ -240,7 +238,7 @@ Here, the `.<number>` settings for the ID and the comment ensure a sufficient wi
 
 Snakemake will check the status of your jobs 40 seconds after submission. Another attempt will be made in 10 seconds, then 20, etcetera with an upper limit of 180 seconds.
 
-## Using Profiles
+### Using Profiles
 
 When using [profiles](https://snakemake.readthedocs.io/en/stable/executing/cli.html#profiles), a command line may become shorter. A sample profile could look like this:
 
@@ -281,7 +279,7 @@ export SNAKEMAKE_PROFILE="$HOME/.config/snakemake"
 
 ==This is ongoing development. Eventually you will be able to annotate different file access patterns.==
 
-## Retries - Or Trying again when a Job failed
+### Retries - Or Trying again when a Job failed
 
 Some cluster jobs may fail. In this case Snakemake can be instructed to try another submit before the entire workflow fails, in this example up to 3 times:
 
@@ -311,15 +309,15 @@ snakemake --slurm-requeue ...
 
 To prevent failures due to faulty parameterization, we can dynamically adjust the runtime behaviour:
 
-## Dynamic Parameterization
+### Dynamic Parameterization
 
 Using dynamic parameterization we can react on different different inputs and prevent our HPC jobs from failing.
 
-### Adjusting Memory Requirements
+#### Adjusting Memory Requirements
 
 Input size of files may vary. [If we have an estimate for the RAM requirement due to varying input file sizes, we can use this to dynamically adjust our jobs.](https://snakemake.readthedocs.io/en/stable/snakefiles/rules.html#dynamic-resources)
 
-### Adjusting Runtime
+#### Adjusting Runtime
 
 Runtime adjustments can be made in a Snakefile:
 
@@ -346,71 +344,21 @@ set-resources:
 Be sure to use sensible settings for your cluster and make use of parallel execution (e.g. threads) and [global profiles](#using-profiles) to avoid I/O contention. 
 
 
-## Nesting Jobs (or Running this Plugin within a Job)
+### Nesting Jobs (or Running this Plugin within a Job)
 
 Some environments provide a shell within a SLURM job, for instance, IDEs started in on-demand context. If Snakemake attempts to use this plugin to spawn jobs on the cluster, this may work just as intended. Or it might not: depending on cluster settings or individual settings, submitted jobs may be ill-parameterized or will not find the right environment.
 
 If the plugin detects to be running within a job, it will therefore issue a warning and stop for 5 seconds.
 
-## Retries - Or Trying again when a Job failed
 
-Some cluster jobs may fail. In this case Snakemake can be instructed to try another submit before the entire workflow fails, in this example up to 3 times:
-
-```console
-snakemake --retries=3
-```
-
-If a workflow fails entirely (e.g. when there are cluster failures), it can be resumed as any other Snakemake workflow:
-
-```console
-snakemake --rerun-incomplete
-```
-
-To prevent failures due to faulty parameterization, we can dynamically adjust the runtime behaviour:
-
-## Dynamic Parameterization
-
-Using dynamic parameterization we can react on different different inputs and prevent our HPC jobs from failing.
-
-### Adjusting Memory Requirements
-
-Input size of files may vary. [If we have an estimate for the RAM requirement due to varying input file sizes, we can use this to dynamically adjust our jobs.](https://snakemake.readthedocs.io/en/stable/snakefiles/rules.html#dynamic-resources)
-
-### Adjusting Runtime
-
-Runtime adjustments can be made in a Snakefile:
-
-```Python
-def get_time(wildcards, attempt):
-    return f"{1 * attempt}h"
-
-rule foo:
-    input: ...
-    output: ...
-    resources:
-        runtime=get_time
-    ...
-```
-
-or in a workflow profile
-
-```YAML
-set-resources:
-    foo:
-        runtime: f"{1 * attempt}h"
-```
-
-Be sure to use sensible settings for your cluster and make use of parallel execution (e.g. threads) and [global profiles](#using-profiles) to avoid I/O contention. 
-
-
-## Summary:
+### Summary:
 
 When put together, a frequent command line looks like:
 
 ```console
 $ snakemake --workflow-profile <path> \
 > -j unlimited \ # assuming an unlimited number of jobs
-> --default-resources slurm_account=<account> slurm_partition=<default partition> \
+> --workflow-profile <profile directory with a `config.yaml`>
 > --configfile config/config.yaml \
-> --directory <path> # assuming a data path not relative to the workflow
+> --directory <path> # assuming a data path on a different file system than the workflow
 ```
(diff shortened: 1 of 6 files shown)


snakemake_executor_plugin_slurm/__init__.py
```diff
@@ -107,6 +109,30 @@ def warn_on_jobcontext(self, done=None):
...
+        self.logger.debug(f"jobs length: {len(jobs)}")
+        # sys.exit(1)
+        # g = groupby(jobs, key=Job.rule)
+        # self.logger.debug(f"Jobs grouped by rule: {g}")
+        # sys.exit(1)
+        for job in jobs:
+            if job.name == "foo":
+                break
+            self.run_job_pre(job)
+            self.run_job(job)
```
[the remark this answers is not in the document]
cmeesters: ah, this should not be part of this PR - this is work in progress and will be submitted in a different PR, once ready. Will be removed when this PR is updated.

snakemake_executor_plugin_slurm/utils.py
```diff
@@ -14,3 +14,21 @@ def delete_slurm_environment():
...
+    for root, _, files in os.walk(logdir, topdown=False):
+        for fname in files:
+            file_path = os.path.join(root, fname)
+            if age_cutoff > 0:
+                filestamp = os.stat(file_path).st_mtime
+                if filestamp > cutoff_secs:
+                    os.remove(file_path)
+        # remove empty rule top dir, if empty
+        if len(os.listdir(root)) == 0:
+            os.rmdir(root)
```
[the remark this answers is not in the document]
cmeesters: accounted for in the latest commits

snakemake_executor_plugin_slurm/utils.py
```diff
@@ -14,3 +14,21 @@
...
+    in the SLURM 'logdir'
+    """
+    cutoff_secs = age_cutoff * 86400
+    for root, _, files in os.walk(logdir, topdown=False):
+        for fname in files:
+            file_path = os.path.join(root, fname)
+            if age_cutoff > 0:
+                filestamp = os.stat(file_path).st_mtime
+                if filestamp > cutoff_secs:
+                    os.remove(file_path)
```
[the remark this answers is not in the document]
cmeesters: the code has been moved: within the function, outside the executor class, the `logger` instance of Snakemake is not accessible.

snakemake_executor_plugin_slurm/utils.py
```diff
@@ -14,3 +14,21 @@
...
+        # remove empty rule top dir, if empty
+        if len(os.listdir(root)) == 0:
+            os.rmdir(root)
```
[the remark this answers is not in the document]
cmeesters: has been corrected

snakemake_executor_plugin_slurm/__init__.py
```diff
@@ -12,6 +12,7 @@
 import time
 from dataclasses import dataclass, field
 from datetime import datetime, timedelta
+import pathlib
```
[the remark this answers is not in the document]
cmeesters: done.

snakemake_executor_plugin_slurm/__init__.py
```diff
@@ -123,18 +196,21 @@ def run_job(self, job: JobExecutorInterface):
...
 
-        slurm_logfile = os.path.abspath(
-            f".snakemake/slurm_logs/{group_or_rule}/{wildcard_str}/%j.log"
+        slurm_logfile = (
+            self.workflow.executor_settings.logdir
+            + os.path.sep
+            + f"{group_or_rule}/{wildcard_str}/%j.log"
         )
-        logdir = os.path.dirname(slurm_logfile)
+
```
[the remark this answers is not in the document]
cmeesters: yep, this is better, is committed. will be in next push.

snakemake_executor_plugin_slurm/__init__.py
```diff
@@ -380,6 +456,11 @@
                     self.report_job_success(j)
                     any_finished = True
                     active_jobs_seen_by_sacct.remove(j.external_jobid)
+                    if not self.workflow.executor_settings.keep_successful_logs:
+                        self.logger.debug(
+                            f"removing log for successful job with SLURM ID '{j.external_jobid}'"
+                        )
+                        os.remove(j.aux["slurm_logfile"])
```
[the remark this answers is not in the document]
cmeesters: done

snakemake_executor_plugin_slurm/__init__.py
```diff
@@ -31,13 +32,51 @@
...
+        metadata={
+            "help": """
+                   Per default the SLURM log directory (writing output is
+                   required by SLURM) is '~/.snakemake/slurm_logs'.
+                   This flag allows to set an alternative directory.
+                   """,
+            "env_var": False,
+            "required": False,
+        },
+    )
```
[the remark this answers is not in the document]
cmeesters: done. This is actually safer because `/home/$USER` might not exist on a cluster. This code was experimental, now relying on the `~`. Is committed. Will be pushed.

snakemake_executor_plugin_slurm/__init__.py
```diff
@@ -123,18 +169,21 @@
         except AttributeError:
             wildcard_str = ""
 
-        slurm_logfile = os.path.abspath(
-            f".snakemake/slurm_logs/{group_or_rule}/{wildcard_str}/%j.log"
+        slurm_logfile = (
+            self.workflow.executor_settings.logdir
+            + os.path.sep
+            + "{log_path}/{group_or_rule}/{wildcard_str}/%j.log"
```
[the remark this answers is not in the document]
cmeesters: done. committed

snakemake_executor_plugin_slurm/__init__.py
```diff
@@ -26,18 +27,53 @@
 from snakemake_interface_common.exceptions import WorkflowError
 from snakemake_executor_plugin_slurm_jobstep import get_cpus_per_task
 
-from .utils import delete_slurm_environment
+from .utils import delete_slurm_environment, clean_old_logs
```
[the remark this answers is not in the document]
cmeesters: done

snakemake_executor_plugin_slurm/__init__.py
```diff
@@ -92,6 +128,13 @@ def __post_init__(self):
         self._fallback_partition = None
         self._preemption_warning = False  # no preemption warning has been issued
 
+    def __del__(self):
+        # the logger attribute is not available in this destructor!
+        clean_old_logs(
+            self.workflow.executor_settings.logdir,
+            self.workflow.executor_settings.delete_logfiles_older_than,
+        )
+
```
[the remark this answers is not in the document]
cmeesters: already done

cmeesters:
@johanneskoester , @dlaehnemann , @fgvieira - I would be happy if at least two of you agreed on the changes in this PR. There are a number of design decisions and coding decisions to consider. 

- I want all SLURM logs of successful jobs to be auto-deleted. These are just too many files. Users can turn this behaviour off.
- Ultimately, all SLURM logs shall be deleted. After 10 days is the default. No reason to keep old SLURM logs: The job logs are, after all, partial Snakemake logs as Snakemake is the job script. And the Snakemake logs are unaffected by this.
- Per default, all SLURM logs will be written to `~/.snakemake/slurm_logs`. This is a constant which allows the previous feature to work. Users may overwrite this behaviour.
- The clean-up method is triggered at exit. For this, the `atexit` module is just. Triggering by `__del__` has its own issues. For `atexit.register()` to work, the clean-up methods needs to proceed the `__post_init__` method. Not nice, but stable.

docs/further.md
```diff
@@ -165,8 +165,12 @@ As always, these can be set on the command line or in a profile.
...
-| `--slurm_requeue` | allows jobs to be resubmitted automatically if they fail or are preempted. See the [section "retries" for details](#retries)|
+| `--slurm-logdir` | can change the default directory for SLURM logs, which is `~/.snakemake/slurm_logs/`|
+| `--slurm-keep-successful-logs` | per default all SLURM logs of successful jobs will be deleted upon job completion, these are just too many files too keep. The flag will overwrite this behaviour.|
+| `--slurm-delete-logfiles-older-than` | all log files in `--slurm-logdir` will be deleted after some time (default is after 10 days). If set to <= 0 this deletion will be skipped. The flag allows to change the default.|
+| `--slurm-init-seconds-before-status-checks`| modify time before initial job status check; the default of 40 seconds avoids load on querying slurm databases, but shorter wait times are for example useful during workflow development |
+| `--slurm-requeue` | allows jobs to be resubmitted automatically if they fail or are preempted. See the [section "retries" for details](#retries)|
```
johanneskoester, line 172: This table should not be put here at all. The Snakemake plugin catalog generates a table with the CLI args automatically: https://snakemake.github.io/snakemake-plugin-catalog/plugins/executor/slurm.html#usage
Hence it is sufficient to put the documentation into the settings dataclass.

snakemake_executor_plugin_slurm/__init__.py
```diff
@@ -83,6 +123,38 @@ class ExecutorSettings(ExecutorSettingsBase):
...
+                    if file_age > cutoff_secs:
+                        os.remove(file_path)
+                except (OSError, FileNotFoundError) as e:
+                    self.logger.warning(f"Could not delete file {file_path}: {e}")
+            # remove empty rule top dir, if empty
+            try:
+                if len(os.listdir(root)) == 0:
+                    os.rmdir(root)
+            except (OSError, FileNotFoundError) as e:
+                self.logger.warning(f"Could not remove empty directory {root}: {e}")
```
johanneskoester, line 156: @coderabbitai, can you suggest a rewrite of this using pathlib?

johanneskoester (changes requested):

docs/further.md
cmeesters: A, good one. Will be corrected in an update of the PR. Certainly helps to deconvolute the docs.

snakemake_executor_plugin_slurm/__init__.py
cmeesters: will be tested

docs/further.md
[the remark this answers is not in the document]
cmeesters: is deleted, according to @johanneskoester's suggestion.

cmeesters: done

snakemake_executor_plugin_slurm/__init__.py
cmeesters: done - with minor adjustments - see last push.

cmeesters:
@johanneskoester care to take a look - seems as if the slurmdbd in the CI is somehow broken, though?

docs/further.md
johanneskoester: For the, the table is still here in the diff.

snakemake_executor_plugin_slurm/utils.py
```diff
@@ -1,6 +1,9 @@
 # utility functions for the SLURM executor plugin
 
 import os
+import logging
+
+logger = logging.getLogger(__name__)
```
johanneskoester, line 6:
```suggestion
```

Not needed  I think. Snakemake has its own logger instance that we access with `self.logger`.

snakemake_executor_plugin_slurm/__init__.py
```diff
@@ -380,6 +452,19 @@ async def check_active_jobs(
                     self.report_job_success(j)
                     any_finished = True
                     active_jobs_seen_by_sacct.remove(j.external_jobid)
+                    if not self.workflow.executor_settings.keep_successful_logs:
+                        self.logger.debug(
+                            f"""removing log for successful job
+                                with SLURM ID '{j.external_jobid}'"""
```
johanneskoester, line 458: Don't use a multiline string here.

snakemake_executor_plugin_slurm/__init__.py
```diff
@@ -104,6 +169,9 @@ def warn_on_jobcontext(self, done=None):
                 delete_slurm_environment()
         done = True
 
+    # def delete_old_logs(self):
+    #    self.workflow.executor_settings.delete_logfiles_older_than
+
```
johanneskoester, line 174: Cleanup?

snakemake_executor_plugin_slurm/__init__.py
```diff
@@ -83,6 +124,28 @@ class ExecutorSettings(ExecutorSettingsBase):
...
+            try:
+                if path.is_file():
+                    file_age = current_time - path.stat().st_mtime
+                    if file_age > cutoff_secs:
+                        path.unlink()
+                    elif path.is_dir() and not any(path.iterdir()):
+                        path.rmdir()
+            except (OSError, FileNotFoundError) as e:
+                self.logger.warning(f"Could not delete file {path}: {e}")
+
```
johanneskoester, line 148: Move below __post_init__ for readability?

snakemake_executor_plugin_slurm/__init__.py
```diff
@@ -47,7 +87,8 @@ class ExecutorSettings(ExecutorSettingsBase):
         metadata={
             "help": """
                     Allow requeuing preempted of failed jobs,
-                    if no cluster default. Results in `sbatch ... --requeue ...`
+                    if no cluster default. Results in
+                    `sbatch ... --requeue ...`
                     This flag has no effect, if not set.
                     """,
```
johanneskoester, line 93: let us not use multiline strings for these, but rather simple strings that are auto-concatenated by python if just separated by whitespace.

snakemake_executor_plugin_slurm/__init__.py
```diff
@@ -31,13 +33,51 @@
...
+        metadata={
+            "help": """
+                   Per default the SLURM log directory (writing output is
+                   required by SLURM) is '~/.snakemake/slurm_logs'.
+                   This flag allows to set an alternative directory.
+                   """,
+            "env_var": False,
+            "required": False,
+        },
+    )
```
johanneskoester, line 47: I think the default should be in the workdir and not in the users home, like it was before.

johanneskoester (changes requested):

johanneskoester:
> @johanneskoester care to take a look - seems as if the slurmdbd in the CI is somehow broken, though?

The action is fixed again. It was using an old slurm plugin that is deprecated in the new slurm release that is on the current ubuntu.

docs/further.md
cmeesters: weird: https://github.com/snakemake/snakemake-executor-plugin-slurm/pull/159/commits/8260f6bb252ded6c7c34f2402bfb3d3b669c2bb3

snakemake_executor_plugin_slurm/__init__.py
cmeesters: This was what I posted to be discussed: By changing the default, the clean-up will be enacted _every_ time. Alternatively, a user has to clean up hidden files, eventually. In my experience, this is what will not happen.

We are talking about SLURM logs, here. Log files, which do not contain substantial information. As Snakemake submits itself, it will mostly have the same information as printed on the terminal — particularly for failed jobs in the case of workflows using the `log` directive, which we ought to expect. The number of log files can be enormous. They may have a direct impact on a file quota in a work space (if enabled). We do not need to keep any in the long run. And we should think about core facilities or users who run different workflows and potentially accumulate this redundant information at different places. 

Granted: Deleting logs of successful runs makes a huge difference, already. Granted: This is a deviation from the current default.

If you still want to keep the current default, I will alter this PR accordingly.

snakemake_executor_plugin_slurm/__init__.py
cmeesters: ah, this was a left-over from the initial tinkering, where I violated the name space, if not before post-init. Will check and change according to the suggestion, when I find the time - ought to be working now.

docs/further.md
cmeesters: not there, even and particularly not after a pull

snakemake_executor_plugin_slurm/__init__.py
cmeesters: done

snakemake_executor_plugin_slurm/__init__.py
cmeesters: done

snakemake_executor_plugin_slurm/__init__.py
cmeesters: done

snakemake_executor_plugin_slurm/__init__.py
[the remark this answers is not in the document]
cmeesters: I have done this myself. Now, the code looks ugly to me. But fine, if this is desired ...

snakemake_executor_plugin_slurm/__init__.py
```diff
@@ -104,6 +169,9 @@
                 delete_slurm_environment()
         done = True
 
+    # def delete_old_logs(self):
+    #    self.workflow.executor_settings.delete_logfiles_older_than
+
```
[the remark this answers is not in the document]
cmeesters: done already, isn't it?

snakemake_executor_plugin_slurm/__init__.py
```diff
@@ -31,13 +33,51 @@
...
+        metadata={
+            "help": """
+                Per default SLURM log files in the SLURM log directory
+                of a workflow will be deleted after 10 days. For this,
+                best leave the default log directory unaltered.
+                Setting this flag allows to change this behaviour.
+                If set to <=0, no old files will be deleted.
+                """
+        },
+    )
```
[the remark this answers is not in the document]
cmeesters: done.

snakemake_executor_plugin_slurm/__init__.py
```diff
@@ -31,13 +33,51 @@
...
+                   Per default SLURM log files will be deleted upon sucessful
+                   completion of a job. Whenever a SLURM job fails, its log
+                   file will be preserved.
+                   This flag allows to keep all SLURM log files, even those
+                   of successful jobs.
+                   """,
+            "env_var": False,
+            "required": False,
+        },
+    )
```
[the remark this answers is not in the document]
cmeesters: done

snakemake_executor_plugin_slurm/__init__.py
```diff
@@ -83,6 +124,28 @@
...
+            try:
+                if path.is_file():
+                    file_age = current_time - path.stat().st_mtime
+                    if file_age > cutoff_secs:
+                        path.unlink()
+                    elif path.is_dir() and not any(path.iterdir()):
+                        path.rmdir()
+            except (OSError, FileNotFoundError) as e:
+                self.logger.warning(f"Could not delete file {path}: {e}")
+
```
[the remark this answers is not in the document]
cmeesters: done with commit https://github.com/snakemake/snakemake-executor-plugin-slurm/pull/159/commits/63b4f59a679fc26cae408f90d0261ed44a5cbf8b

cmeesters:
@johanneskoester Happy New Year!

IMO, it's just the issue with the default path, which might be subject to change. Please take a look and weight the argument.

snakemake_executor_plugin_slurm/__init__.py
johanneskoester: I don't yet get why having logs in the user directory makes a difference in terms of clean-up. Can you explain that again? Sorry if I miss something here. I would think that cleanup can work in the same way if things are located in the workdir.

In general, I think that the logs should live in the workdir by default. Having them in the user dir can even cause conflicts caused by multiple instances of the same workflow, or at least make things harder to associate.

cmeesters: Ideally, a workflow is invoked per workdir only once. If everything worked successfully, no SLURM logfile will be left. Else, with several failed attempts, there will be a number of failed log files, which will not be cleaned up. The user will run the next workflow (the same on different data or a different one) in a different workdir. This will take us from the opportunity to conduct an automated cleanup.

When the default (is left unchanged and) points to a user's home, it is a constant. Hence, automated cleanup over time will kick in. With different workdirs the automated cleanup after X days, can be deleted again. It would be a rather useless feature.

Name clashes cannot occur, as each SLURM logfile carries the ` JOB_ID`.

BTW, I found the SLURM logfiles of failed jobs utterly bereft of information: They only contain additional information, if the `log` directive is missing in the respective rule. Else, it is the same information as printed on the terminal. Snakemake will, for the most part, only report an error case, not the cause.

Again, I am fine with either way. If you prefer to keep the default as-is, I will change the code accordingly.

johanneskoester: Ok, I understand the problem now. Still, it feels better to keep the log in the workdir by default. Maybe the help text can mention that cleanup is even more effective if all logs are written to a common directory (because of the reason you mentioned). Then, admins who provide a profile can also make that decision for the user.

cmeesters: In my experience, admins will not bother. They will leave the users to their own devices and are still baffled when dealing with workflow engines.

Anyhow. I am going to change the PR accordingly.

snakemake_executor_plugin_slurm/__init__.py
cmeesters: done

snakemake_executor_plugin_slurm/__init__.py
cmeesters: done

snakemake_executor_plugin_slurm/utils.py
```diff
@@ -14,3 +15,37 @@ def delete_slurm_environment():
...
+        done = True
+        for p in dirs:
+            try:
+                if not any(p.iterdir()):  # if directory is empty
+                    p.rmdir()
+            except (OSError, FileNotFoundError) as e:
+                raise e  # re-raise exception
+            finally:
+                done = False
+                dirs.remove(p)
```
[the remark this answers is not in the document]
cmeesters: @coderabbit: fine, will commit something similar, but it appears, this code this does not do what I want: to erase _all_ empty directories recursively. The recursive bit is still missing, hence the suggested change, will be in the next commit.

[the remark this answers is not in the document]
cmeesters: by all means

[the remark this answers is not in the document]
cmeesters: @coderabbitai - yes, a suggestion would be nice to have. By now, I have a "good enough solution", only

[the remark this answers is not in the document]
cmeesters: I admit, this is an elegant 101 level solution to the problem. Thanks!

cmeesters:
@johanneskoester all changes are in - I added docs, too. To me, everything is now good enough. Some refactoring is due, soon.

snakemake_executor_plugin_slurm/__init__.py
```diff
@@ -26,30 +28,59 @@
 from snakemake_interface_common.exceptions import WorkflowError
 from snakemake_executor_plugin_slurm_jobstep import get_cpus_per_task
 
-from .utils import delete_slurm_environment
+from .utils import delete_slurm_environment, delete_empty_dirs
 
 
 @dataclass
 class ExecutorSettings(ExecutorSettingsBase):
+    logdir: Optional[str] = field(
```
johanneskoester, line 36:
```suggestion
    logdir: Optional[Path] = field(
```

snakemake_executor_plugin_slurm/__init__.py
```diff
@@ -380,6 +443,19 @@ async def check_active_jobs(
...
+                            "removing log for successful job "
+                            f"with SLURM ID '{j.external_jobid}'"
+                        )
+                        try:
+                            if j.aux["slurm_logfile"].exists():
+                                j.aux["slurm_logfile"].unlink()
+                        except (OSError, FileNotFoundError) as e:
+                            self.logger.warning(
+                                "Could not remove log file"
+                                f" {j.aux['slurm_logfile']._str}: {e}"
```
johanneskoester, line 457:
```suggestion
                                f" {j.aux['slurm_logfile']}: {e}"
```

johanneskoester (changes requested):

johanneskoester (approved):

PR #211, opened 2025-02-14
fix: missing quotes for the slurm comment

benjamin-lieser:
Snakemake does not correctly escape the comment it passes to slurm. When a wildcard contains for example a `(` this will lead to errors.

## Summary by CodeRabbit

Minor internal updates have been deployed to streamline how job submission commands are assembled. These refinements maintain the current functionality and user experience without introducing any visible changes.

- **Refactor**
  - Enhanced internal handling of command formatting during job submissions to ensure consistency.

### DIFF
--- a/snakemake_executor_plugin_slurm/__init__.py
+++ b/snakemake_executor_plugin_slurm/__init__.py
@@ -208,7 +208,7 @@ class Executor(RemoteExecutor):
             f"--job-name {self.run_uuid} "
             f"--output '{slurm_logfile}' "
             f"--export=ALL "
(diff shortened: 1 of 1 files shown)


cmeesters:
Thank you for your contribution @benjamin-lieser ! Does not seem like a big deal. However, I like to ask: Has there been an issue for you? Can you give a bit of context?

cmeesters (approved):
minor change - will prevent fails in case of syntax containing characters within wildcards

cmeesters:
@benjamin-lieser thank you for your contribution!

PR #231, opened 2025-03-12
ci: mastodon action

cmeesters:
this PR attempts to utilize the new mastodon-github-action release.


.github/workflows/announce-release.yml
```diff
@@ -0,0 +1,35 @@
+name: Announce Release on Mastodon
+
+on:
+  pull_request:
+    types:
+      - closed
+    branches:
+      - main
```
dlaehnemann, line 8: I think this is the better way to target merges into the main branch, and to restrict the workflow to only run when there have been changes to the `CHANGELOG.md`:
```suggestion
  push:
    branches:
      - main
    path:
      - 'CHANGELOG.md'
```
Here's some background on this:
https://stackoverflow.com/a/61565445

And I would still leave in the check `chore(main): release` check for the actual job, to avoid toots when someone manually edits the `CHANGELOG.md`.

.github/workflows/announce-release.yml
```diff
@@ -0,0 +1,35 @@
...
+
+permissions:
+  pull-requests: read
+
+jobs:
+  post_to_mastodon:
+    if: |
+      github.event.pull_request.merged == true &&
+      (contains(github.event.pull_request.title, 'release') || contains(github.event.pull_request.title, ' v') || contains(github.event.pull_request.title, 'version'))
```
dlaehnemann, line 17: Maybe make this more specific, locking it to the exact title that release-please creates:
```suggestion
    if: contains(github.event.pull_request.title, 'chore(main): release')
```
Also, I think the `pull_request_title` never contains a `v` or a `version`, right? And with the different targeting of the workflow I suggested above, I think there is also no need for the `github.event.pull_request.merged` check.

dlaehnemann (commented):
Looks very nice, two suggestions on the workflow triggering / targeting.

dlaehnemann (approved):
:+1:

PR #237, opened 2025-03-13
docs: review and edit new docs

dlaehnemann:
## Summary by CodeRabbit

- **Documentation**
  - Updated `docs/further.md` with extensive restructuring and detailed guidance on installation, resource configuration, MPI, GPU, local execution, advanced settings, and job nesting.
  - Added comprehensive examples, best practices, and troubleshooting tips for SLURM cluster interaction and resource management.

- **Chores**
  - Updated the release announcement workflow to use a newer version of the Mastodon release post action.

### DIFF
--- a/docs/further.md
+++ b/docs/further.md
@@ -39,8 +39,8 @@ $ snakemake --executor slurm \
 > --directory <path>
 ```
 
-This examples assumes no limit for submitted jobs (`-j unlimited`). Any number, e.g. `-j 150`, will throttle a workflow to this number of concurrent jobs.
-Furtheremore we on many clusters we have to assume seperation from workflows and data, hence the flag `--directory <path>` which can point to a different file system, containing your data.
+This example assumes no limit for submitted jobs (`-j unlimited`). Any number, e.g. `-j 150`, will throttle a workflow to this number of concurrent jobs.
+Furthermore, on many clusters we must assume separation of workflows and data. Use the `--directory <path>` flag to point to a file system that contains your data.
 
 ### Configuration
 


cmeesters:
@dlaehnemann please mark the PR as 'ready for review' once done.

dlaehnemann:
Jup, but currently I'm still working on this. Hope to get through it today!

cmeesters:
@dlaehnemann hi, updated the branch a couple of times. So far, only minor updates. I would like to update main from this branch soon-ish as some work needs documentation. (I'm on it.) Yet, I understand that your changes are substantial.

cmeesters:
@dlaehnemann sample profiles have been deleted alltogether. Admittingly, the docs profit from scraping my verbose text. Yet, a sample configuration yaml along with the encouragement to contribute examples to existing workflows and best practise describtion would greatly benefit users.

I will write a suggestion.

dlaehnemann:
> @dlaehnemann yes, the main documentation covers MPI. Frankly, this documentation is even more in need of an update. And the main documentation should point to the HPC-Executors rather than the other way around: The only sensible environment to start MPI-Jobs are clusters, and you require those executor plugins. People will look for "their" plugin and its docs. Hence, the executor docs should cover all parallelisation paradigms.
> 
I agree that they should definitely both cover MPI, with the executor docs covering the specifics for an executor, but the main docs should explain the general idea and whatever snakemake itself (and potentially the snakemake-interface-executor-plugins) does. From what I can see, the general mechanism to use MPI is simply to provide all the necessary info under the `resources:` directive and then reference this directly in the `shell:` directive. And I would expect this to be the same mechanism for all executors, so this should probably stay in the main docs? At the same time, any specific recommendations, like `srun` for Slurm, should be in the respective executor docs. And then we should have clear links both ways.

> The default MPI-Executor for SLURM is `srun`. On LSF you need to use generic starters like `mpiexec` or `mpihydra`. The SLURM-Executor only ensures that only rules with a `shell` directive can utilize a mpi-resource. Next, the MPI-process is started from the Jobstep-Executor. Yes, its basic.
So from what I understand, the only Slurm-specific info that is relevant from the user perspective, is to use `srun` as the `resources: mpi=` entry, right? Or am I missing something? And could you point me to the jobstep-executor, so I can understand this better?

Also, there is this code testing for the presence of `tasks_per_node`, but I don't see `tasks_per_node` actually being used anywhere in the code:
https://github.com/snakemake/snakemake-executor-plugin-slurm/blob/bbbcec4f58aa869406a3a3194f6730b88c7ea4df/snakemake_executor_plugin_slurm/__init__.py#L273
So even if users specify this, I guess this would just be ignored, right? And the info message would be misleading? Or did I simply miss the respective code when searching for it?

dlaehnemann:
Finally, I could also go about testing out MPI a bit on our slurm instance. Do you have a good basic test case you could point me to?

cmeesters:
I think, now this is fine. Except for a cluster example for the storage fs plugin, which I deem important. And a comprehensive example for `--profile` vs. `--workflow-profile` in a cluster environment.

cmeesters:
@dlaehnemann about MPI: I removed the TODO item in my latest commit (see https://github.com/snakemake/snakemake-executor-plugin-slurm/pull/237/commits/5686272fca0e4ae7fe7c7e92d26dc41b1a720250 ). Yes, the executor does nothing special, but I want to keep certain redundancy in order to avoid confusion. Are you fine with that?

cmeesters:
@dlaehnemann I removed the TODO item in the requeue section and redrafted a bit (actually, it read pretty well). see https://github.com/snakemake/snakemake-executor-plugin-slurm/pull/237/commits/80073654390a5f15c900bd21018410f5ca10c626

cmeesters:
@dlaehnemann I removed the TODO items from the status check section (see: https://github.com/snakemake/snakemake-executor-plugin-slurm/pull/237/commits/4cfac882e5488041897c4ed5932f2003c5b9fdc8):
the plugin uses the build-in rate limiter (see the main code, in this pr line 400 `async with self.status_rate_limiter:`), this obeys the `--seconds-between-status-checks` and `--max-status-checks-per-second` options, does it not?

Also, I do not want to give recommendations - to me the defaults plus the own code are fine, and I have no idea of other admin's attitudes. Together with @johanneskoester I discussed these parameters a lot (back in the day before the plugin) and yes, users can limit checks further. So far, there has not been an issue or feature request regarding this topic.

What do you think? Leave it like this?

cmeesters (approved):
To me, this looks good now. I will, however, wait for a second review.

Also, I want to add more examples in subsequent PRs.

cmeesters:
@dlaehnemann please have a look at the latest version. Is it too your liking? What needs to be addressed?

docs/further.md
```diff
@@ -7,103 +7,204 @@ To avoid redundancy, the plugin deletes the SLURM log file for successful jobs,
...
-``` console
-$ snakemake --executor slurm --default-resources slurm_account=<your SLURM account> slurm_partition=<your SLURM partition>
+A command line invocation of the plugin could look like:
+
+```console
+$ snakemake --executor slurm \
+> -j unlimited \ # assuming an unlimited number of jobs
+> --workflow-profile <profile directory with a `config.yaml`>
+> --configfile config/config.yaml \
+> --directory <path> # assuming a data path on a different file system than the workflow
```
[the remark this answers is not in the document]
cmeesters: @coderabbitai - corrected in the latest commit.

docs/further.md
```diff
@@ -180,38 +177,34 @@ You can check whether your cluster has this enabled with:
...
-If enabled, this feature allows jobs to be automatically resubmitted if they fail or are preempted.
-This effectively does not consider the SLURM job failed, preserving job IDs and priorities, and allowing priority to be accumulated while pending.
-If job requeuing is not enabled on your cluster, consider adding `--slurm-requeue` for your Snakemake jobs:
+This should show a numerical value for the `JobRequeue` parameter, indicating the number of requeues for individual jobs.  
+Any number greater than 1 will cause SLURM to requeue jobs on your cluster upon failure or preemption this many times. (Your cluster may specify additional parameters.)
 
-TODO: Does the `--slurm-requeue` command line argument work when this feature is on, or when it is off. This is not clear from the previous docs version.
+If enabled, this feature allows jobs to be automatically resubmitted if they fail or are preempted - you do not need to rely on this plugin in this case.
+
+If your cluster does not support automated requeing, you can ask the plugin to requeu preemted jobs with the `--slurm-requeue` flag:
```
dlaehnemann, line 186:
```suggestion
If your cluster does not support automated requeuing, you can ask the plugin to requeue preempted jobs with the `--slurm-requeue` flag: 
```

docs/further.md
```diff
@@ -180,38 +177,34 @@ You can check whether your cluster has this enabled with:
...
+#### MPI-specific Resources
 
 Snakemake's SLURM executor plugin supports the execution of MPI ([Message Passing Interface](https://en.wikipedia.org/wiki/Message_Passing_Interface)) jobs.
-Per default, jobs can only run on a single cluster node (or machine) and parallelization is thus limited by the maximum number of cores that is available on any machine in the cluster.
-MPI jobs enable parallel computations spanning across multiple nodes, thus potentially parallelizing to more cores than any machine in you cluster can offer.
+
+Per default, jobs can only run on a single cluster node (or machine) and parallelization is thus limited by the maximum number of cores that is available on any machine in the cluster. MPI jobs enable parallel computations spanning across multiple nodes, thus potentially parallelizing to more cores than any machine in you cluster can offer. [See the main documentation section of Snakemake, too.](https://snakemake.readthedocs.io/en/stable/snakefiles/rules.html#mpi-support)
```
dlaehnemann, line 199: I would stick with one line per sentence:
```suggestion
Per default, jobs can only run on a single cluster node (or machine) and parallelization is thus limited by the maximum number of cores that is available on any machine in the cluster.
MPI jobs enable parallel computations spanning across multiple nodes, thus potentially parallelizing to more cores than any machine in you cluster can offer.
[See the main documentation section of Snakemake, too.](https://snakemake.readthedocs.io/en/stable/snakefiles/rules.html#mpi-support)
```

docs/further.md
```diff
@@ -180,38 +177,34 @@ You can check whether your cluster has this enabled with:
...
 
 | Snakemake     | Description                                 |
 |---------------|---------------------------------------------|
 | `mpi`         | MPI launcher command, for example `srun`    |
-| `tasks`       | the walltime per job in minutes             |
+| `tasks`       | The number of SLURM tasks - equivalent to MPI-ranks |
 
 
-To effectively utilize MPI within a Snakemake workflow, it's recommended to use `srun` as the MPI launcher when operating in a SLURM environment.
+To effectively utilize MPI within a Snakemake workflow, it's recommended to use `srun` as the MPI launcher when operating in a SLURM environment. However, some programs do not work well with this MPI launcer or require a detailed topology layout - this can be added to the `srun` statement, if required.
```
dlaehnemann, line 207: Again, one line per sentence for better diffs:
```suggestion
To effectively utilize MPI within a Snakemake workflow, it's recommended to use `srun` as the MPI launcher when operating in a SLURM environment.
However, some programs do not work well with this MPI launcer or require a detailed topology layout - this can be added to the `srun` statement, if required.
```

docs/further.md
```diff
@@ -385,7 +378,7 @@ By defining these resource specifications in a profile, you maintain a clean and
 ### Running Jobs locally
 
 In Snakemake workflows executed within cluster environments, certain tasks -- such as brief data downloads or plotting -- are better suited for local execution on the head node rather than being submitted as cluster jobs.
-To designate specific rules for local execution, Snakemake offers the localrules directive.
+To designate specific rules for local execution, Snakemake offers the `localrules` directive. For more details, refer to the [Snakemake documentation on local rules](https://snakemake.readthedocs.io/en/stable/snakefiles/rules.html#local-rules).
```
dlaehnemann, line 381: Again, one line per sentence for better diffs:
```suggestion
To designate specific rules for local execution, Snakemake offers the `localrules` directive.
For more details, refer to the [Snakemake documentation on local rules](https://snakemake.readthedocs.io/en/stable/snakefiles/rules.html#local-rules).
```

docs/further.md
```diff
@@ -157,17 +157,14 @@ Some of these are [central snakemake command-line arguments determining its beha
   To reduce the burden of regular status checks,  this time automatically increases in 10 second steps with every status check whenever no finished jobs are found.
   But this will never go beyond a wait time of 180 seconds (3 minutes), to avoid long wait times once jobs do finish.
   Also, as soon as finished jobs are found, this gets reset to 40 seconds.
-  (TODO: respect the `--seconds-between-status-checks` option in the plugin and use it as the minimum wait time.)
```
dlaehnemann: Should we add a new issue to track this TODO? Or have you addressed this, already?

docs/further.md
```diff
@@ -157,17 +157,14 @@ Some of these are [central snakemake command-line arguments determining its beha
...
 * With `--max-status-checks-per-second`, you can limit the frequency of individual attempts of querying the job status database within a round of status checks.
   Within a round, repeated attempts will happen whenever the `sacct` command used for the status query comes back with an error.
-  (TODO: set a reasonable default here; also double-check this is what we want to happen, here.)
```
dlaehnemann: As with the TODO above, should we track this in a new issue? Maybe both can go into one issue to thoroughly review the wait times / frequencies settings in the plugin.

docs/further.md
```diff
@@ -157,17 +157,14 @@ Some of these are [central snakemake command-line arguments determining its beha
...
 
 Or you might sometimes want to decrease certain wait times for small workflow runs during development.
 For example, the plugin waits 40 seconds before performing the first job status check.
-You can reduce this with the `--slurm-init-seconds-before-status-checks=<time in seconds>` option, to minimize turn-around times for test runs.
-TODO: add wait times and frequencies arguments to SLURM-specific table
```
dlaehnemann: Have you decided against putting those parameters in the SLURM-specific table?

dlaehnemann (commented):
These are basically just formatting suggestions and questions if we maybe want to track some deleted TODOs in an issue. Otherwise, I think this is fine to merge.

I had been hoping to find time to test mpi and gpu integration on our cluster, but this is just not happening any time soon. But if I can find the time, I'll simply revisit the docs (and possibly the plugin code) accordingly!

docs/further.md
dlaehnemann: Ah wait, just saw your earlier comment. Will have another look, first.

dlaehnemann:
> @dlaehnemann I removed the TODO items from the status check section (see: [4cfac88](https://github.com/snakemake/snakemake-executor-plugin-slurm/commit/4cfac882e5488041897c4ed5932f2003c5b9fdc8)): the plugin uses the build-in rate limiter (see the main code, in this pr line 400 `async with self.status_rate_limiter:`), this obeys the `--seconds-between-status-checks` and `--max-status-checks-per-second` options, does it not?

That `status_rate_limiter` just uses the `--max-status-checks-per-second` option, see here:
https://github.com/snakemake/snakemake-interface-executor-plugins/blob/bfbe3e4650b97c2fd868f36b740f55cd7c8c1bcb/snakemake_interface_executor_plugins/executors/remote.py#L90

And this will only limit how many `sacct` commands are issued per second within a round of status check attempts. So the default for this should probably rather be some lower fraction like 0.2, to make it wait 5 seconds before re-attempting the `sacct`.

But from what I can see in the snakemake code, the default is actually `100.0`:
https://github.com/snakemake/snakemake/blob/91cc3ef0f24b1b0e96cb751cdc806cad84491425/src/snakemake/settings/types.py#L420

Or `10`, if we trust the rendered CLI  docs:
https://snakemake.readthedocs.io/en/stable/executing/cli.html#snakemake.cli-get_argument_parser-behavior

Both of these would be rapid firing `sacct` at the slurm system as many times as you attempt a status check. So maybe instead put something like `max_status_checks_per_second=0.2` here:
https://github.com/snakemake/snakemake-executor-plugin-slurm/blob/bb42d3934c77eab0c745459056af9c18bdf70e27/snakemake_executor_plugin_slurm/__init__.py#L130


As for the `--seconds-between-status-checks`, I think this is initialized here:
https://github.com/snakemake/snakemake-interface-executor-plugins/blob/bfbe3e4650b97c2fd868f36b740f55cd7c8c1bcb/snakemake_interface_executor_plugins/executors/remote.py#L280

And we update it here, resetting it to 40 seconds whenever any jobs are finished:
https://github.com/snakemake/snakemake-executor-plugin-slurm/blob/bb42d3934c77eab0c745459056af9c18bdf70e27/snakemake_executor_plugin_slurm/__init__.py#L502

So I think we should at least reset it to the command-line value. And we should double-check that simply setting this value will have the intended effect of waiting. Because I don't understand the code calling this in the interface. There's first the sleep function, that uses this:
https://github.com/snakemake/snakemake-interface-executor-plugins/blob/bfbe3e4650b97c2fd868f36b740f55cd7c8c1bcb/snakemake_interface_executor_plugins/executors/remote.py#L269

And this sleep function is in turn only used here:
https://github.com/snakemake/snakemake-interface-executor-plugins/blob/bfbe3e4650b97c2fd868f36b740f55cd7c8c1bcb/snakemake_interface_executor_plugins/executors/remote.py#L191

And eventually it is used here, within the `threading.Thread()` call:
https://github.com/snakemake/snakemake-interface-executor-plugins/blob/bfbe3e4650b97c2fd868f36b740f55cd7c8c1bcb/snakemake_interface_executor_plugins/executors/remote.py#L83

So I guess this somehow uses the `threading` core python module to wait for jobs and update the active_jobs list, but I can't seem to wrap my head around how it does this. But it probably works as intended and I simply don't understand it... :sweat_smile:

docs/further.md
dlaehnemann: OK, this is now in the discussion below, so I'll resolve here.

docs/further.md
dlaehnemann: OK, this is now in the discussion below, so I'll resolve here.

dlaehnemann:
As stated above, maybe this is best tracked in a separate issue. We could just copy over my verbose discussion there and then finally merge this PR...?!

cmeesters:
> As stated above, maybe this is best tracked in a separate issue. We could just copy over my verbose discussion there and then finally merge this PR?

Yes, just one more grammar update and your opinion on the SLURM flags table. As for the separate issue - my idea is that the rate limiter takes both flags into account. And I would not know how to propagate plugin information as `max_status_checks_per_second=0.2` back. Honestly, I am confused. I never observed that many status checks as suggested by docs and code. I need to inspect.

dlaehnemann:
> > As stated above, maybe this is best tracked in a separate issue. We could just copy over my verbose discussion there and then finally merge this PR?
> 
> Yes, just one more grammar update and your opinion on the SLURM flags table. As for the separate issue - my idea is that the rate limiter takes both flags into account. And I would not know how to propagate plugin information as `max_status_checks_per_second=0.2` back. Honestly, I am confused. I never observed that many status checks as suggested by docs and code. I need to inspect.

Sorry for the slow reaction. I was on vacation after my last comments/review. But good to see this merged and I added a new issue for us to follow up on the waiting times configurations: #319

Also see [my comment on the table question](https://github.com/snakemake/snakemake-executor-plugin-slurm/pull/237#discussion_r2120561558). What do you think?

PR #260, opened 2025-05-07
fix: Increase account charecter limit

ChristofferCOASD:
I faced an issue where a long queue name is prepended with a +, which causes the test_account to fail. I recommend increasing the charecter.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved validation of SLURM accounts by updating the formatting of the account check command to ensure more consistent and reliable results.


cmeesters:
Thanks for this contribution!

cmeesters (approved):
A trivial change improving cross-cluster portability.

PR #262, opened 2025-05-08
Sync with upstream release v1.2.99

gwct:
Auto-created sync PR for upstream version v1.2.99


gwct:
Sorry, this was not meant to make a PR back to this repo. Please delete.

cmeesters:
@gwct I am happy to welcome PRs to the SLURM executor (and am open to video calls to discussing issues beforehand, if this deems necessary), but consider it a bad idea to have all kind of specific SLURM executors for specific clusters: Every new plugin will eventually deviate from the main SLURM executor and be out of sync. This will include all updates to the main SLURM executor.

gwct:
@cmeesters Sorry about this. The plugin for our Cannon cluster is not meant to be pulled back into the main SLURM plugin. It will be an independent plugin that is a fork of the SLURM plugin that we maintain and keep in sync with the main plugin. I definitely agree that you would never want cluster specific logic in the main SLURM plugin.

This PR was the result of a test of an auto-sync and PR workflow action that mistakenly opened this PR back to the original plugin, which was a mistake. Apologies again for cluttering your PR history.

PR #285, opened 2025-05-15
Enables running snakemake job within a slurm cluster reservation.

bmilash:
This change lets the plugin start a slurm job using a reservation resource provided within a profile or on the command line.

### DIFF
--- a/snakemake_executor_plugin_slurm/submit_string.py
+++ b/snakemake_executor_plugin_slurm/submit_string.py
@@ -50,6 +50,11 @@ def get_submit_command(job, params):
     if job.resources.get("nodes", False):
         call += f" --nodes={job.resources.get('nodes', 1)}"
 
+    # Specify reservation for job if any. Reservation name is in
+    # single quotes since reservation names may include spaces.
+    if job.resources.get("reservation"):
+        call += f" --reservation='{job.resources.reservation}'"
+
     # fixes #40 - set ntasks regardless of mpi, because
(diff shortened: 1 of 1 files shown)


cmeesters:
Thank you for your contribution!

Honestly? I thought about that flag, too. Then I learned about the `magnetic` flag for reservations and as I only use reservations during courses, adding something to the executor did not seem necessary any longer. Is there a need for this flag?

snakemake_executor_plugin_slurm/submit_string.py
```diff
@@ -50,6 +50,11 @@ def get_submit_command(job, params):
     if job.resources.get("nodes", False):
         call += f" --nodes={job.resources.get('nodes', 1)}"
 
+    # Specify reservation for job if any. Reservation name is in
```
cmeesters, line 53: is this is a resource per job, users would need to alter workflows or write a default resource in their workflow profiles. Both scenarios not really desirable, when wanting to stay as generic as possible. See the comment thread for alternatives.

bmilash: Hi Christian,
I hadn't considered relying on the magnetic attribute of a reservation. I agree, this would be best as a default resource for a workflow, or passed on the command line, and not a per job resource. This capability is something we would use during post-downtime cluster testing. I'll discuss use of magnetic reservations with our sys admins.
Thanks,
Brett

Issue #320, opened 2025-06-04
`SLURM sbatch failed` causes Snakemake crash


nikostr:
**Software Versions**

snakemake:  8.27.1
snakemake-executor-plugin: 1.3.6
SLURM: 24.05.4

**Describe the bug**
`sbatch: error: Batch job submission failed: Requested node configuration is not available` leads to Snakemake crashing with `WorkflowError`, with jobs being left running/in the SLURM queue.

A minimal improvement would be for the workflow to shut down nicely, either canceling jobs and cleaning up after them, or waiting for them to finish. Another option would be to treat this like other job failures, allowing to workflow to keep going, and even allowing retries for jobs failing due to this. Also, specifying which exact call to SLURM caused the error would help with troubleshooting these issues.

**Logs**

The full error message:
```
WorkflowError:
SLURM sbatch failed. The error message was sbatch: error: Batch job submission failed: Requested node configuration is not available
``` 

**Additional context**
My cluster's resource limits aren't super clear. Some combination of resources generated in later attempts are clearly not okay.

cmeesters:
Thank you for your issue report. Please note:
- the current Snakemake version is 9.5.1 . Yours is a bit outdated. During the Snakemake Hackathon in March we made a number of fixes potentially impacting cluster execution, too.
- However, your particular error indicates, that you requested something, which your cluster does not provide. Please check with `snakemake [<args>] --verbose`. There you see the actual `sbatch` call made by the SLURM executor.

nikostr:
No worries, we all have those days :) To be clear, this isn't urgent :)

I have already updated my snakemake version.

I'm not sure that I'm following? When an sbatch call fails a `WorkflowError` is raised, if I understand the following code correctly:
https://github.com/snakemake/snakemake-executor-plugin-slurm/blob/ec82a70b95652afcbc21d44a4dbca14d78dc8936/snakemake_executor_plugin_slurm/__init__.py#L288-L304

Since this leads to the workflow immediately crashing when I run this with the latest version of snakemake I'm assuming that this isn't being caught and handled anywhere, but I may very well misunderstand something.

nikostr:
Oh, and regarding the Snakefile above: you might need to tweak the sleep times depending on the queues etc. For me these values worked. What needs to happen is that `gen_b` needs to be pending or running when `gen_c` is submitted and crashes, leaving `gen_b` in the queue without it being handled by snakemake.

nikostr:
[...]

PR #322, opened 2025-06-09
feat: treat sbatch errors as job errors instead of workflow errors

nikostr:
fixes https://github.com/snakemake/snakemake-executor-plugin-slurm/issues/320
When I do some very rudimentary testing it seems to work: when sbatch fails, it is handled like a job error rather than a workflow error.

The argument for `report_job_error` needs to be a `SubmittedJobInfo`. These jobs are technically never reported as submitted. I'm not sure if that would cause some type of problem? Would this PR need some sort of tests included?

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved error handling during SLURM job submission to prevent unnecessary interruptions when an error occurs, ensuring smoother workflow execution.
  - Fixed minor formatting issue related to job status checking for more accurate time calculations.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -19,7 +19,7 @@ snakemake-interface-common = "^1.13.0"
 snakemake-interface-executor-plugins = "^9.1.1"
 snakemake-executor-plugin-slurm-jobstep = "^0.3.0"
 pandas = "^2.2.3"
-numpy = "^1.26.4"
+numpy = ">=1.26.4, <3"
 throttler = "^1.2.2"
 
 [tool.poetry.group.dev.dependencies]


johanneskoester (approved):
Good catch! A proper test would be very complex I think, including mocking and so on. For me it would be fine to live without a test for now.

No submission report is fine, there is no need to do that before the error.

In general, I think it is good to cancel the executor upon any unforeseen error, such that no unmanaged jobs remain behind. Hence, I have created a corresponding PR: https://github.com/snakemake/snakemake/pull/3660

johanneskoester:
Now just the linting fails.

nikostr:
I think I have resolved the linting issue. A lint in an unrelated part of the code also got fixed. Additionally I also sorted out the formatting of the messages, now it looks fairly consistent with the rest of the info being displayed.

I've noticed that the message does not get displayed when errors are reported again when the workflow completes, but this issue seems to be unrelated to this plugin.

cmeesters:
@nikostr & @johanneskoester 

Hi guys,

I just updated this branch from main. nikostr, I asked Johannes to have a look, for a was a bit lost. Now, with the remark on the Snakemake core update, it makes a lot more sense.

If you like, Johannes, proceed with merging and releasing. I will be back in ~ 3 weeks time.

Cheers,
Christian

PR #330, opened 2025-07-08
feat: add prefix to job uuid

sebhoerl:


snakemake_executor_plugin_slurm/__init__.py
```diff
@@ -148,6 +148,14 @@ class ExecutorSettings(ExecutorSettingsBase):
             "required": False,
         },
     )
+    prefix: str = field(
```
cmeesters, line 151: "prefix" is a bit too generic. How about naming the purpose, e.g. "prefix_uuid"?

cmeesters (changes requested):
Most importantly: I would like to know the reason for this PR. Some motivation.

PR #336, opened 2025-08-06
feat: make status and cancel command configurable

ebirn:
Slurms behavior with sacct and scancel (with specific options) changed when no accounting database is configured:

```
[supercomputer ~]# sacct 
Slurm accounting storage is disabled

[supercomputer ~]# scancel 123 --clusters=all
scancel: error: Problem talking to database
scancel: error: No clusters can be reached now. Contact your admin to resolve the problem.
scancel: fatal: Could not get cluster information
```

My Slurm version in use:
```
# scancel --version
slurm 24.05.4
```

## Slurm job status

For checking the job status an aequivalent commant to the current implementation can be run with squeue (given the Slurm config has sufficiently long, i.e. `MinJobAge=86400`).

Example an alternative status command:
```
$ squeue --format=%i\|%T --states=all --noheader --name b722e167-0559-4add-8c32-2fcc7aa80a62
3266|CANCELLED
3267|CANCELLED
[...]
```

## slurm job cancel
For the cancel command, I'd rather remove the `--clusters=all` flag, and make the default just to issue the cancel on the "local" cluster (with the option to override it). But I'm not an active Snakemake user, so the discussion about default settings is better left to someone else ;)

### DIFF
--- a/snakemake_executor_plugin_slurm/__init__.py
+++ b/snakemake_executor_plugin_slurm/__init__.py
@@ -158,8 +158,8 @@ class ExecutorSettings(ExecutorSettingsBase):
             "help": "The command to query the status of SLURM jobs. "
             "This command should return one line for each job with "
             "<raw/main_job_id>|<long_status_string>."
-            "If no accounting is enabled, an alternative is:"
-            "squeue --states=all --format=%i\|%T --noheader --name {run_uuid}",
+            "If no accounting is enabled, an alternative is:",
+            # "squeue --states=all --format=%i|%T --noheader --name {run_uuid}",
             "env_var": False,
             "required": False,
         },
@@ -614,7 +614,7 @@ We leave it to SLURM to resume your job(s)"""
                 # about 30 sec, but can be longer in extreme cases.
                 # Under 'normal' circumstances, 'scancel' is executed in
                 # virtually no time.
-                scancel_command = self.workflow.executor_settings.canceland.format(jobids=jobids)
+                scancel_command = self.workflow.executor_settings.cancel_command.format(jobids=jobids)
 
                 subprocess.check_output(
                     scancel_command,


cmeesters:
I consider a cluster without accounting db as not completely set up. Hence, there is no assumption, that `sacct` will work without an accounting db, but rather that some sort of accounting, albeit volatile, can easily be set up and is a rather sensible thing to do - for admins and users alike.

- that `squeue` reliably displays completed jobs is news to me, I must admit. Will have to test (we are a year behind your SLURM version and the plugin has to work for a range of versions).
- if `squeue` worked reliably in this regard, I would welcome an easy, (perhaps) preconfigured, switch. Wouldn't you agree to have a range of preconfigured options better in terms of abstracting complexity and ease of use?

We would, however, need a test suite to do that. Are you familiar with the test framework we use? Else we might have to schedule a meeting, no necessarily with you (but welcoming you, if you like) for this is not an easy task. I'm afraid, your case is not yet covered by our framework.

ebirn:
Large clusters will have accounting enabled, but we provide Slurm also on smaller application specific appliances, where accounting is not an issue, but it is used for ordering access to resources to i.e. GPUs.

For the switch, you imagine something like a `--no-sacct` switch, that switches to squeue for status and calls scancel without the --clusters=all option?

I'm not familiar with the test framework, sorry.

cmeesters:
> For the switch, you imagine something like a

Something along the line - yes. Have to think about it. I am open for suggestions.

> Large clusters will have accounting enabled ...

My argument was always, that without, there is information missing for admins and users alike. If the missing functionality is indeed now present for `squeue`, we can have that switch. I have to admit: I do wonder how retrieving the job status for completed jobs (longer than a minute or two in the past) can be accomplished without a DB. But I cannot test this myself - it does not work on our clusters. Can you confirm, it does for your cluster?

I will attempt a PR with test cases. We would need to update your PR accordingly.

I understand the urgency in this case. (Although setting up SLURM with an accounting DB is rather straightforward.) Please understand that I am under teaching obligations and currently without a cluster to test on (it is being moved). So, no promises as to timely solutions.

As for the `scancel` issue: This is related but different. How about a separate PR (I can do this), introducing a switch for multiclusters? In this case, the default would be without `--all` and no DB query can fail.

I am however reluctant to introduce more and more switches: the UI should not be too obfuscated. So, whenever you have an ideain that regard, let me know about it, please.

cmeesters:
I have been working on a test. However, without a cluster (should be up tomorrow 🤞), I cannot test a thing, and it is a mess right now, so no commit.

As to point 2: I think we need to query this parameterization to make the choice available. After all, this _is_ volatile memory and I assume it will not be configured at all to some substantial time on many clusters. What is your take on a minimum time which is sensible? My guess for long-running workflows it ought to be 12 h minimum. This is because, a workflow can easily run for days in extreme scenarios and jobs need to be queried. On the other side, a bigger cluster will not cache such information as there may be too much information to store (considering the number of jobs - and a user might cause OOM issues for the ctld if a permutation test goes wrong). So, my bet would be to query this parameterization and make one or both options available, based on the presence of commands (leaning precedence to `sacct`) and offer the choice only, if both options are available and reasonably save. What do you think?

Please understand that we need to deploy this as stable as possible. I am glad, you found a working solution and appreciate your PR!

PR #345, opened 2025-08-21
fix: trying test speedup

cmeesters:
just an attempt to speed up the CI


cmeesters:
did not work as expected

PR #375, opened 2025-11-14
fix: extracting job id from convoluted output, if necessary

cmeesters:
The job-ID extraction only worked if `--parsable` output is not overwritten by cluster administrators. This fix attempts a parsing of such output, knowing, that it might fail, if too many integers are in this output.

## Summary by CodeRabbit

* **Bug Fixes**
  * Submission now normalizes/recovers SLURM job IDs from scheduler output to reduce false failures and provide clearer errors for missing/ambiguous IDs.

* **New Features**
  * Added a reusable job-ID validation utility.
  * Added a robust time-to-seconds parser for SLURM time formats and integrated it into reporting.

* **Tests**
  * Extensive new tests covering job-ID extraction edge cases, time parsing, and reporting/parsing behavior.

* **Chores**
  * Minor formatting cleanups and a small formatter configuration tweak.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/tests/tests.py
+++ b/tests/tests.py
@@ -7,7 +7,7 @@ from snakemake_interface_executor_plugins.settings import ExecutorSettingsBase
 from unittest.mock import MagicMock, patch
 import pytest
 from snakemake_executor_plugin_slurm import ExecutorSettings
-from snakemake_executor_plugin_slurm.utils import set_gres_string, time
+from snakemake_executor_plugin_slurm.utils import set_gres_string
 from snakemake_executor_plugin_slurm.submit_string import get_submit_command
 from snakemake_executor_plugin_slurm.validation import (
     validate_slurm_extra,


nateybear:
Amazing! Thanks for providing such a quick fix! I am going to leave a few comments in the code changes. Excellent work.

snakemake_executor_plugin_slurm/validation.py
```diff
@@ -6,6 +6,45 @@
...
+    else:
+        # Try matching a positive integer, raise an error if more than one match or
+        # no match found. Match standalone integers, excluding those followed by %, letters,
+        # or digits (units/percentages/floats). Allows format: "1234" or
+        # "1234; clustername" (SLURM multi-cluster format).
+
+        # If the first attempt to validate the job fails, try parsing the sbatch output
+        # a bit more sophisticatedly:
+        matches = re.findall(
+            r"\b\d+(?![%A-Za-z\d.])", output
```
nateybear: What are we capturing here? Word boundary, then some digits, then capturing something that isn't a digit or character? This may be from misunderstanding the regex but I would think we want to capture `\d+`.

cmeesters: Look at your ouput: There is a xx.y%. Capturing all integers would yiel: `[<jobid>, xx, y]`. That's the thing: Admins may write whatever they want to add to the sbatch output. It may contain several numbers. We have no good reason to assume, the jobid will always be the last number. We have no good reason to assume it is the only parsable number. The regex attempts to discard all numbers followed immediately by a `%` sign, a period, or letters. I am open to suggestions.

I refrained from checking all parsable numbers against `squeue` as that would display a tremendous overhead on the SLURM control daemon. That would, however, be a secure way to validate the job IDs.

cmeesters: I probably should put a verbose comment there. It's not, that I will understand the regex right away in a stressful situation or should expect the co-developers will do ...

[will be off-screen now, have a nice weekend]

snakemake_executor_plugin_slurm/__init__.py
```diff
@@ -382,8 +382,10 @@ def run_job(self, job: JobExecutorInterface):
         # To extract the job id we split by semicolon and take the first element
         # (this also works if no cluster name was provided)
         slurm_jobid = out.strip().split(";")[0]
+        # this slurm_jobid might be wrong: some cluster admin give convulted
+        # sbatch outputs. So we need to validate it properly:
         if not slurm_jobid:
-            raise WorkflowError("Failed to retrieve SLURM job ID from sbatch output.")
+            slurm_jobid = validate_slurm_job_id(slurm_jobid, out)
```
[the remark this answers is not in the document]
cmeesters: done

snakemake_executor_plugin_slurm/validation.py
cmeesters: done

cmeesters:
@johanneskoester merging the new main resulted in issues. Will continue tomorrow.

PR #396, opened 2025-12-22
fix: status hang when no active jobs

kjohnsen:
When running from a SLURM job (I know this is discouraged...) somehow `check_active_jobs()` was called before there were any active jobs, which made it hang (it never hit a yield or return statement). This fixed it for me

## Summary by CodeRabbit

* **Refactor**
  * Optimized job status checking to skip unnecessary queries when no jobs are active
  * Made status check intervals configurable for more flexible resource management

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/snakemake_executor_plugin_slurm/__init__.py
+++ b/snakemake_executor_plugin_slurm/__init__.py
@@ -666,11 +666,6 @@ class Executor(RemoteExecutor):
         # decide which status command to use
         status_command_name = self.get_status_command()
         min_job_age = get_min_job_age()
-        initial_interval = getattr(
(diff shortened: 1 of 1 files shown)


kjohnsen:
Yes, I had the exact same Executor `run_uuid` problem co-occur with this issue. I noticed it went away when I ran with the debugger (since it was slower, maybe that gave the Executor time to configure, get run_uuid?)

cmeesters (approved):
The change seems minor with regard to functionality. But might change a user experience drastically.

PR #448, opened 2026-04-02
fix: handle integer slurm_account values from YAML parsing

andchamorro:
## Summary

Snakemake's YAML parser automatically converts numeric-looking strings (e.g., `"123456789"`) to integers when populating `job.resources`. This caused `account.lower()` in `test_account()` to fail since `int` has no `lower()` method.

## Changes

Convert `slurm_account` values to strings before use in `re.split()` to handle both string and integer values.

## Tests

Added 3 new unit tests covering:
- String account values
- Integer account values (from YAML parsing)
- Multiple string accounts

## Summary by CodeRabbit

* **Bug Fixes**
  * Normalize SLURM account inputs (including numeric values) to strings, split multiple accounts, and ensure submission arguments are consistently formatted and quoted.

* **Tests**
  * Added unit tests validating single, numeric, and multi-account scenarios to ensure correct account argument generation and invocation.

### DIFF
--- a/snakemake_executor_plugin_slurm/__init__.py
+++ b/snakemake_executor_plugin_slurm/__init__.py
@@ -1293,7 +1293,7 @@ class Executor(RemoteExecutor):
         # slurmdbd/accounting hiccups where job states are briefly incomplete.
         for i in range(status_attempts):
             async with self.status_rate_limiter:
-                (status_of_jobs, sacct_query_duration) = await query_job_status(
+                status_of_jobs, sacct_query_duration = await query_job_status(
                     status_command, self.logger
                 )
                 if status_of_jobs is None and sacct_query_duration is None:
@@ -1589,8 +1589,10 @@ We leave it to SLURM to resume your job(s)"""
         if job.resources.get("slurm_account"):
             # split the account upon ',' and whitespace, to allow
             # multiple accounts being given
+            # Note: YAML parsing may convert numeric strings (e.g. "123456") to int.
+            # Ensure we always work with strings for re.split and shlex.quote.
             accounts = [
-                a for a in re.split(r"[,\s]+", job.resources.slurm_account) if a
+                a for a in re.split(r"[,\s]+", str(job.resources.slurm_account)) if a
             ]
             for account in accounts:
                 # here, we check whether the given or guessed account is valid
--- a/tests/test_account.py
+++ b/tests/test_account.py
@@ -1,0 +1,81 @@
+"""Unit tests for get_account_arg() method."""
+
+from unittest.mock import MagicMock, patch
+
+from snakemake_executor_plugin_slurm import Executor
+
+
+class _Resources(dict):
+    """Dict-like resources with attribute access for known keys only."""
+
+    def __getattr__(self, name):
+        try:
+            return self[name]
+        except KeyError as e:
+            raise AttributeError(name) from e
+
+
+def _make_mock_job(
+    rule_name="myrule", name=None, wildcards=None, jobid=1, is_group=False, **resources
+):
+    """Return a minimal mock job compatible with get_account_arg."""
+    mock_resources = _Resources(resources)
+
+    mock_rule = MagicMock()
+    mock_rule.name = rule_name
+
+    job = MagicMock()
+    job.resources = mock_resources
+    job.rule = mock_rule
+    job.name = name if name is not None else rule_name
+    job.wildcards = wildcards if wildcards is not None else {}
+    job.is_group.return_value = is_group
+    job.threads = resources.get("threads", 1)
+    job.jobid = jobid
+    return job
+
+
+class TestGetAccountArg:
+    """Tests for get_account_arg() method handling string and integer accounts."""
+
+    def _make_executor_stub(self):
+        """Return a minimal Executor stub."""
+        executor = Executor.__new__(Executor)
+        executor.logger = MagicMock()
+        executor._fallback_account_arg = None
+        return executor
+
+    # Patches prevent actual SLURM command execution (sacct/sacctmgr).
+    # - get_account() would call 'sacct' to guess account from recent jobs.
+    # - test_account() would call 'sacctmgr' or 'sshare' to validate account.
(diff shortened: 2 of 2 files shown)


snakemake_executor_plugin_slurm/__init__.py
```diff
@@ -1602,15 +1602,15 @@ def get_account_arg(self, job: JobExecutorInterface):
             # contain build-in shell commands - see issue #354
             for account in accounts:
                 test_account(account, self.logger)
-                yield f" -A {shlex.quote(account)}"
+                yield f" -A {shlex.quote(str(account))}"
```
cmeesters: is this conversion necessary, if you convert an account to str before?

snakemake_executor_plugin_slurm/__init__.py
```diff
@@ -1602,15 +1602,15 @@ def get_account_arg(self, job: JobExecutorInterface):
...
         else:
             if self._fallback_account_arg is None:
                 self.logger.warning("No SLURM account given, trying to guess.")
                 account = get_account(self.logger)
                 if account:
                     self.logger.warning(f"Guessed SLURM account: {account}")
                     test_account(f"{account}", self.logger)
-                    self._fallback_account_arg = f" -A {shlex.quote(account)}"
+                    self._fallback_account_arg = f" -A {shlex.quote(str(account))}"
```
cmeesters: is this conversion necessary, after the conversion which happened already?

tests/test_array_jobs.py
```diff
@@ -641,3 +641,52 @@ def test_non_terminal_status_keeps_job_active(self, monkeypatch, tmp_path):
         assert remaining == [active_job]
         executor.report_job_success.assert_not_called()
         executor.report_job_error.assert_not_called()
+
+
+class TestGetAccountArg:
```
cmeesters: account testing ought not to go into the array test script

tests/test_array_jobs.py
```diff
@@ -641,3 +641,52 @@ def test_non_terminal_status_keeps_job_active(self, monkeypatch, tmp_path):
...
+    """Unit tests for get_account_arg() method."""
+
+    def _make_executor_stub(self):
+        """Return a minimal Executor stub."""
+        executor = Executor.__new__(Executor)
+        executor.logger = MagicMock()
+        executor._fallback_account_arg = None
+        return executor
+
+    @patch("snakemake_executor_plugin_slurm.get_account")
```
cmeesters: why these patches?

cmeesters (commented):
Thank you for your contribution! I left a couple of comments to address. Generally speaking, I am happy to accept.

snakemake_executor_plugin_slurm/__init__.py
andchamorro: It isn't specifically necessary for the situation I encountered; it's a preventive measure suggested by Sonnet.

snakemake_executor_plugin_slurm/__init__.py
andchamorro: It isn't specifically necessary for the situation I encountered; it's a preventive measure suggested by Sonnet.

tests/test_array_jobs.py
andchamorro: unittest mock patch decorator that replace get_account() and test_account(). Replace the real function with MagicMock objects.

tests/test_array_jobs.py
andchamorro: My mistake, I just associate array with slurm arrays. I will move to test/test_account.py

tests/test_array_jobs.py
cmeesters: ah yes. Mind leaving a comment? Or I would stumble across this and accidentally remove it sometime into the future.

snakemake_executor_plugin_slurm/__init__.py
cmeesters: I understand the reason, but would rather have a comment about the type (I will introduce more and more type annotations to functions) and where it has been set than redundant conversions.

tests/test_array_jobs.py
andchamorro: I will leave a comment in the new test location tests/test_account.py.

snakemake_executor_plugin_slurm/__init__.py
andchamorro: I agree, I will remove the edits and leave some comments.

tests/test_array_jobs.py
cmeesters: done

tests/test_account.py
```diff
@@ -0,0 +1,83 @@
...
+        assert results == [" -A acc1", " -A acc2", " -A acc3"]
+
+    @patch("snakemake_executor_plugin_slurm.get_account")
+    @patch("snakemake_executor_plugin_slurm.test_account")
+    def test_multiple_accounts_integer(self, mock_test_account, mock_get_account, tmp_path):
+        """Multiple integer accounts work correctly."""
+        executor = self._make_executor_stub()
+        job = _make_mock_job(slurm_account=123)
+        results = list(executor.get_account_arg(job))
+        assert results == [" -A 123"]
```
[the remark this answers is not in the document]
cmeesters: @coderabbitai: that is correct. Yet, the support for multiple accounts is anticipated, but not actually present. Hence, we there is no need to change this, now.

snakemake_executor_plugin_slurm/__init__.py
cmeesters: Sorry, was on a holiday after Easter. The code looks fine. I do not see the commit which changes it, but that might be an artefact.

cmeesters:
@andchamorro I was on a holiday after Easter. Actually, I think, the PR is ready to be merged. Please correct the formatting issue (Just run `black` on the code.)

andchamorro:
I hope you had a wonderful holidays. I've already made the necessary formatting changes. Thanks for everything.

cmeesters (approved):

PR #450, opened 2026-04-05
fix: pipeline hangs when submitting from compute nodes

jayhesselberth:
When running snakemake from within a SLURM job (e.g., an interactive session on a compute node), the pipeline would submit jobs but never detect their completion, hanging forever.

The RemoteExecutor base class starts a status-checking daemon thread in `__init__` *before* `__post_init__` is called. The SLURM plugin's `warn_on_jobcontext()` in `__post_init__` would sleep 5 seconds and then delete SLURM environment variables, but by then the daemon thread had already started and would silently die after its first polling cycle.

Fix: move the SLURM environment detection and cleanup into `__init__`, before `super().__init__()` starts the daemon thread. Remove the now unnecessary `warn_on_jobcontext()` method and its 5-second sleep.

## Summary by CodeRabbit

* **Bug Fixes**
  * Cleaner SLURM environment detection and immediate cleanup during executor startup, with an earlier warning when a SLURM job context is present to improve job submission reliability.
* **Tests**
  * Test suite updated to align with the revised executor initialization and warning behavior; expectations remain unchanged.

### DIFF
--- a/snakemake_executor_plugin_slurm/__init__.py
+++ b/snakemake_executor_plugin_slurm/__init__.py
@@ -428,9 +428,22 @@ def _select_logdir(workflow):
 # Required:
 # Implementation of your executor
 class Executor(RemoteExecutor):
+    def __init__(self, workflow, logger):
+        # Must clean SLURM environment BEFORE super().__init__() starts
+        # the wait_thread, otherwise the thread races with __post_init__
+        # and silently dies, causing the pipeline to hang forever.
+        if "SLURM_JOB_ID" in os.environ:
+            logger.warning(
+                "You are running snakemake in a SLURM job context. "
+                "This is not recommended, as it may lead to unexpected "
+                "behavior. "
+                "If possible, please run Snakemake directly on the "
+                "login node."
+            )
+            delete_slurm_environment()
+        super().__init__(workflow, logger)
+
     def __post_init__(self, test_mode: bool = False):
-        # run check whether we are running in a SLURM job context
-        self.warn_on_jobcontext()
         self.test_mode = test_mode
 
         self.run_uuid = str(uuid.uuid4())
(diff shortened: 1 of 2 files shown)


cmeesters:
Thanks for this PR!

At the Snakemake Hackathon I noticed, that even when unsetting all `$SLURM...` env vars _before_ starting Snakemake within a job, all jobs are submitted with only one thread. I did not find the time to investigate. Are you experiencing the same issue? If not, what did you do? Perhaps we can profit from that experience.

cmeesters:
@jayhesselberth I am actually fine with this PR. Will you apply `black` on the code to fix the formatting issue? 

What I meant by my last remark: If you have an order of commands which solves the start-within-jobcontext-issue, I am eager to learn.

jayhesselberth:
@cmeesters in our case, it was a combination of this fix and not having `sacct` set up correctly on some of our compute nodes (some couldn't talk to slurmdbctl).

cmeesters (approved):
@jayhesselberth Ok, I will fix the formatting prior to the next release, but will merge it already.

gernophil:
While this definitely seems to make it more robust, if started from a compute node, I had my workflow hanging today for 4.5h without submitting any jobs before I had to manually cancel, unlock and restart it:
```
...
[Wed Apr 29 03:49:49 2026]
Finished jobid: 60 (Rule: xxx)
2 of 234 steps (1%) done
Select jobs to execute...
Execute 16 jobs...
[2026-04-29T08:24:21.031] error: *** JOB XXXXXXXX ON hpc-cpu-xxx CANCELLED AT 2026-04-29T08:24:21 DUE to SIGNAL Terminated ***
```

cmeesters:
@gernophil I'm afraid, this is probably a different issue. Please write a comprehensive description as a stand-alone issue report.

gernophil:
@cmeesters Thanks for the feedback. Might be related to wofklow profile vs. defining resources via CLI. I'll perform some more systematic tests and also talk to our Slurm admins and then come back with a stand-alone issue ones I know more.

PR #451, opened 2026-04-10
fix: code refactoring

fgvieira:
- allow for both GRES and GPU to be specified (for example, `gpu` with some network specs)
- assume default GPU of 1
- GRES can be just a string since, according to docs, slurm will assume 1:

**--gres=<list>**
    Specifies a comma-delimited list of generic consumable resources requested per node. The format for each entry in the list is "name[[:type]:count]". The name is the type of consumable resource (e.g. gpu). The type is an optional classification for the resource (e.g. a100). The count is the number of those resources with a default value of 1. [...]. Examples of use include "--gres=gpu:2", "--gres=gpu:kepler:2", and "--gres=help". 

## Summary by CodeRabbit

* **Bug Fixes**
  * Broadened validation for generic resource specs to accept optional subtypes and optional numeric/postfix formats.
  * GPU allocation now consistently appears in job submissions and permits a model identifier without requiring an explicit GPU count (defaults to 1 when omitted).
  * Removed strict mutual-exclusion between generic-resource and GPU settings so combined configurations are allowed.

* **Tests**
  * Updated tests to accept expanded GRES/GPU formats and auto-filled GPU counts.

### DIFF
--- a/snakemake_executor_plugin_slurm/utils.py
+++ b/snakemake_executor_plugin_slurm/utils.py
@@ -341,8 +341,8 @@ def set_gres_string(job: JobExecutorInterface) -> str:
     Function to set the gres string for the SLURM job
     based on the resources requested in the job.
     """
-    # generic resources (GRES) arguments can be of type
-    # "string", "string:int" or "string:string:int" with optional postfix 'T' or 'G' or 'M'
+    # generic resources (GRES) arguments can be of type "string",
+    # "string:int" or "string:string:int" with optional postfix 'T' or 'G' or 'M'
     gres_re = re.compile(r"^[a-zA-Z0-9_]+(:[a-zA-Z0-9_\.]+)?(:\d+[TGM]?)?$")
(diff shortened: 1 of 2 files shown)


fgvieira:
@cmeesters do these changes make sense?

cmeesters:
@fgvieira I don't know. I am currently in Wales and will need some time next week, before I can have a look.

cmeesters:
@fgvieira ok, I agree, this is much cleaner. Plus: Yes, `--gres` might be a necessity in addition to `--gpus` on some clusters. On the other hand, it _can_ result in conflicts.

We should:
- go through the failing tests, which still assume that `--gres` and `--gpus` are mutually exclusive.
- and look for a better error message upon failed job submissions. I fear the latter is hard and beyond the scope of this PR. Perhaps worth a talk to SchedMD.

cmeesters (approved):

PR #468, opened 2026-06-08
feat: remote stage in

cmeesters:
This PR will add the option for multi-node stage in on SLURM clusters - will only work with a corresponding update of jobstep plugin.

The ultimate goal is to enable pooling of shared memory tasks (e.g. all mapping rules of a workflow within one SLURM job, possibly spanning multiple nodes). This is useful, because:
- demanded by some cluster admins, who otherwise see too many jobs
- we then only have to have _one_ stage-in process instead of many (as are now demanded per job by this PR for this plugin or for the fs storage plugin in general).

However, together with the corresponding PR for the jobstep plugin (https://github.com/snakemake/snakemake-executor-plugin-slurm-jobstep/pull/48) this has the additional charm that there is no dependency for the fs storage plugin.

Furthermore, pooling of smp tasks is hardly beneficial if there is no automatic stage-in process as this will inevitably lead to random I/O (which already is an issue as some users do not throttle the submission rate, jobs reading shared reference file and then simple cause random access and impede file system performance).

After this PR, the plugin will use `sbcast` for files <= 4G for multi node stage-in and `scp` otherwise. The latter relies on `ssh` being possible across nodes on a cluster (either by auth based logins on own jobs or otherwise). Currently, there is no fallback to `sbcast` if files are too big and `ssh` is not working.
 

## Summary by CodeRabbit

- **New Features**
  - Added automatic node-local staging for SLURM jobs using access-pattern annotations.
  - Added configuration to set a node-local storage location or disable automatic stage-in.
  - Added size-aware staging behavior, using efficient transfer methods for smaller and larger files.

- **Documentation**
  - Documented configuration, performance benefits, transfer methods, and current limitations.

- **Tests**
  - Added coverage for environment-variable handling and stage-in workflows using multiple transfer mechanisms.


cmeesters:
Good, only the dependency on the jobstep executor is not working - needs a release, first.

Now:
- docs
- and test for working ssh. Perhaps check for a running job whether ssh is possible to a node and give an error when appropriate?

cmeesters:
@johanneskoester & @fbartusch first, the corresponding jobstep PR needs merging for the tests to be succeeding (or be refined as nessary).

.github/workflows/ci.yml
```diff
@@ -19,6 +19,12 @@ jobs:
           cache: true
           cache-write: ${{ github.event_name == 'push' && github.ref_name == 'main' }}
 
+      - name: Debug black environment
+        continue-on-error: true
+        run: |
+          poetry run black --version
+          poetry run black --check --diff --target-version py311 .
```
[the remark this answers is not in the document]
cmeesters: @coderabbitai - done. I removed this redundant check based on black.

johanneskoester (commented):
Sorry, maybe we discussed this already, but why exactly would one use this instead of the fs plugin?

cmeesters:
@johanneskoester 

Because:

1. we support cross-node stage-in. This cannot be accomplished with the fs-plugin unless we support `sbcast` or `scp` in that plugin. But I am longing for cross-node stage-in for the next feature: pooling smp jobs across the node boundary. Which I will not start until we have cross-node stage-in.
2. we drop the necessity to install a plugin not everybody is aware of.

Plus, in my previous position, I had some difficulties with that plugin, that it did not behave well with every workflow. That might have been the result of a cluster quirk. Here, if a local scratch is defined and the annotation is given, it is guaranteed to work. I tested it for multinode jobs, already.

snakemake_executor_plugin_slurm/__init__.py
```diff
@@ -241,6 +247,32 @@ class ExecutorSettings(ExecutorSettingsBase):
         },
     )
 
+    node_local_prefix: Optional[str] = field(
+        default=None,
+        metadata={
+            "help": "A cluster specific node local local global directory. "
+            "Similar to '.remote_job_local_storage_prefix' of the fs-storage "
```
johanneskoester: This prefix is for all storage plugins though. It is not specific for fs-storage.

snakemake_executor_plugin_slurm/__init__.py
```diff
@@ -658,9 +690,20 @@ def additional_general_args(self):
...
+        node_local_prefix = getattr(
+            self.workflow.executor_settings, "node_local_prefix", None
+        )
+        # switch for passing a command as script vs using `sbatch --wrap`
         if self.workflow.executor_settings.pass_command_as_script:
             general_args += " --slurm-jobstep-pass-command-as-script"
+        # attempt auto stage-in if not suppressed and node_local_prefix is set
+        if node_local_prefix and not suppress_auto_stagein:
+            ld = node_local_prefix
+            ld = encode_deferred_envvars(ld)
```
johanneskoester, line 708: I use base64 encoding/decoding in the main snakemake for such args, in order to avoid (theoretical but unlikely) ambiguity caused by the masking of the variable characters. Up to you whether you want to do the same here.

johanneskoester (changes requested):

johanneskoester:
> @johanneskoester
> 
> Because:
> 
>     1. we support cross-node stage-in. This cannot be accomplished with the fs-plugin unless we support `sbcast` or `scp` in that plugin. But I am longing for cross-node stage-in for the next feature: pooling smp jobs across the node boundary. Which I will not start until we have cross-node stage-in.
> 
>     2. we drop the necessity to install a plugin not everybody is aware of.
> 
> 
> Plus, in my previous position, I had some difficulties with that plugin, that it did not behave well with every workflow. That might have been the result of a cluster quirk. Here, if a local scratch is defined and the annotation is given, it is guaranteed to work. I tested it for multinode jobs, already.

Ok, understood. That makes sense. A related question: what happens if an input file uses a remote storage (e.g. S3) and a random access pattern? I think in such a case the slurm plugin based stage-in should not happen, because the job will download the remote storage, right?

Another thought: you should tell people in the docs that they should probably set the remote-job-local-prefix to the same dir as the slurm specific stage in directory to ensure that also remote storage files are handled equally well?

snakemake_executor_plugin_slurm/__init__.py
cmeesters: Addressed by commit [632a845](https://github.com/snakemake/snakemake-executor-plugin-slurm/pull/468/commits/632a84574b65da1e1333fa4b212553225076158b). Thanks for spotting.

PR #470, opened 2026-06-11
fix: Circumvent problematic Slurm retry behaviour

### LINKED ISSUE #348: OOM handling for jobstep plugin
**Software Versions**
<!--Note the Snakemake version and the plugin version for which you experience the bug.
Note the SLURM version, too. You can get the versions with:
$ snakemake --version
9.9.0
$ conda list | grep "snakemake-executor-plugin-slurm"
snakemake-executor-plugin-slurm 1.6.0              pyhdfd78af_0    bioconda
snakemake-executor-plugin-slurm-jobstep 0.3.0              pyhdfd78af_0    bioconda
$ sinfo --version
slurm 24.11.

**Describe the bug**
WHen the slurm_jobstep hits OOM, it keeps retrying even though it hit OOM. I would expect that the system would be resubmitted with more RAM.

**Logs**
```
~# sacct -j [JOBID]
JobID           JobName  Partition    Account  AllocCPUS      State ExitCode
------------ ---------- ---------- ---------- ---------- ---------- --------
[JOBID]      [JOBNAME]-+      debug    default          1    RUNNING      0:0
[JOBID].bat+      batch               default          1    RUNNING      0:0
[JOBID].0    python3.12               default          1 OUT_OF_ME+    0:125
[JOBID].1    python3.12               default          1 OUT_OF_ME+    0:125
[JOBID].2    python3.12               default          1 OUT_OF_ME+    0:125
[JOBID].3    python3.12               default          1 OUT_OF_ME+    0:125
[JOBID].4    python3.12               default          1 OUT_OF_ME+    0:125
[JOBID].5    python3.12               default          1 OUT_OF_ME+    0:125
[JOBID].6    python3.12               default          1 OUT_OF_ME+    0:125
[JOBID].7    python3.12               default          1 OUT_OF_ME+    0:125
[JOBID].8    python3.12               default          1 OUT_OF_ME+    0:125
[JOBID].9    python3.12               default          1 OUT_OF_ME+    0:125
```

**Minimal example**
```
from snakemake.shell import shell

rule use_ram:
    input:
      input = '/dev/zero'
    threads: 1
    resources:
      cores=1,
      mem_mb=10240,
      partition='debug',
      nodes=1
    retries: 10
    shell:
        "head -c 12G    {input} | tail"
```

**Additional context**
What I would expect is that given it runs into OOM, it should just cancel the sbatch or return to the starting snakemake job to potentially get more RAM.

selten:
Tested to fix https://github.com/snakemake/snakemake-executor-plugin-slurm/issues/348 combined with another PR for jobstep plugin. This code was generated with help from AI.

## Summary by CodeRabbit

* **Bug Fixes**
  * Ensured Snakemake job retry/attempt numbers are properly propagated to submitted SLURM jobs, including array jobs and nested job steps.
  * Added reliable cleanup so attempt state is reset after job submission.
* **Tests**
  * Added automated coverage for attempt inheritance and reset behavior.
  * Verified that attempt propagation does not rely on global environment variables.

### DIFF
--- a/snakemake_executor_plugin_slurm/__init__.py
+++ b/snakemake_executor_plugin_slurm/__init__.py
@@ -782,7 +782,7 @@ class Executor(RemoteExecutor):
         try:
             # Set inherited attempt for nested executor (use first job's attempt)
             self.workflow.executor_settings._inherited_attempt = jobs[0].attempt
-            
+
             self.logger.debug(
                 f"Preparing to submit array job for rule {jobs[0].rule.name} "
                 f"with {len(jobs)} tasks."
@@ -1060,7 +1060,7 @@ class Executor(RemoteExecutor):
         try:
             # Set inherited attempt for nested executor
             self.workflow.executor_settings._inherited_attempt = job.attempt
-            
+
             group_or_rule = (
                 f"group_{job.name}" if job.is_group() else f"rule_{job.name}"
             )
--- a/snakemake_executor_plugin_slurm/__init__.py
+++ b/snakemake_executor_plugin_slurm/__init__.py
@@ -855,10 +855,10 @@ class Executor(RemoteExecutor):
                     "your cluster."
                 )
             # Build a compressed map of array task id -> full execution string
-            # for all jobs.
+            # for all jobs. Export attempt state for each job.
             array_execs = {
                 index: zlib.compress(
-                    self.format_job_exec(job).encode("utf-8"), level=9
+                    f"export SNAKEMAKE_ATTEMPT={job.attempt}; {self.format_job_exec(job)}".encode("utf-8"), level=9
                 ).hex()
                 for index, job in enumerate(jobs, start=1)
             }
@@ -885,7 +885,9 @@ class Executor(RemoteExecutor):
                 end_index = min(start_index + array_limit - 1, len(jobs))
                 # The first task of each chunk runs via the plain base command.
                 # Remaining tasks are dispatched from --slurm-jobstep-array-execs.
-                exec_job = self.format_job_exec(jobs[start_index - 1])
+                # Export attempt state for the first task as well.
+                first_job = jobs[start_index - 1]
+                exec_job = f"export SNAKEMAKE_ATTEMPT={first_job.attempt}; {self.format_job_exec(first_job)}"
                 sub_array_execs = {
                     str(i): array_execs[i]
                     for i in range(start_index + 1, end_index + 1)
--- a/snakemake_executor_plugin_slurm/__init__.py
+++ b/snakemake_executor_plugin_slurm/__init__.py
@@ -636,7 +636,8 @@ class Executor(RemoteExecutor):
         passed to `exec_job`.
         """
         general_args = "--executor slurm-jobstep --jobs 1"
-        # need to pass
+        # Pass attempt state to nested executor for resource calculation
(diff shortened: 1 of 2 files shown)


cmeesters:
Thank you for your contribution!

I will need some time to test, as I am busy these days. But I will try to do a preliminary review understanding, that proper tests will fail until the jobstep executor PR is merged, too.

snakemake_executor_plugin_slurm/__init__.py
```diff
@@ -636,7 +636,8 @@ def additional_general_args(self):
         passed to `exec_job`.
         """
         general_args = "--executor slurm-jobstep --jobs 1"
-        # need to pass
+        # Pass attempt state to nested executor for resource calculation
+        general_args += " --envvars SNAKEMAKE_ATTEMPT"
```
cmeesters: I think we should refrain from creating and exporting new environment variables. We introduced an extra environment variable to be picked up on the login node for the partition configuration. But here, we create one on the fly and this is not transparent to users.

Introducing a command line flag for the jobstep executor, like the one used in lines 641 is very much preferred. CLI flags appear in help messages. For env vars we need to ensure in all future development, that there is no clash. While we export all env vars to the SLURM jobs, we cannot always ensure that and will be working on a fix for in-job submissions, which might mess with the environment.

snakemake_executor_plugin_slurm/__init__.py
```diff
@@ -1121,6 +1124,8 @@ def run_job(self, job: JobExecutorInterface):
                 )
 
             exec_job = self.format_job_exec(job)
+            # Export attempt state for nested executor
+            exec_job = f"export SNAKEMAKE_ATTEMPT={job.attempt}; {exec_job}"
```
cmeesters: we should, by no means, tinker with `exec_job` if at all possible. I did this myself for the array jobs, but an exception should remain an exception. The CLI approach would not cause any problems in future developments.

cmeesters (commented):
It would be nice to have a test added to the test suite. But I understand, if this is a bit challenging. Without, however, a release (considering the upcoming holidays) might take a bit longer.

snakemake_executor_plugin_slurm/__init__.py
selten: I completely updated the approach to avoid doing that now.

snakemake_executor_plugin_slurm/__init__.py
```diff
@@ -165,6 +165,17 @@ def _status_lookup_ids(external_jobid: str) -> List[str]:
 class ExecutorSettings(ExecutorSettingsBase):
     """Settings for the SLURM executor plugin."""
 
+    _inherited_attempt: int = field(
```
cmeesters, line 168: I'm afraid, this will translate `--slurm--inherited-attempt`. The leading underscore does not hide an option.

I want to have hidden options, only for internal use, myself.

selten: I do not have an idea how I can accomplish this, do you have an idea? One suggestion that's not ideal would be to set `help=argparse.SUPPRESS`.

cmeesters: No, we should leave it without the underscore for the time being, and I will do a refactoring, once this feature is in the new CLI base.

Sorry for not getting back to you, I was on holiday and changed position. Just settling in, now.

PR #472, opened 2026-06-17
refactor: ci using pixi and ruff/flake8

cmeesters:
the situation with black and python version mismatches on local and CI environments is not bearable any longer. Hence, the switch to a CI supporting pixi, ruff and flake8

## Summary by CodeRabbit

* **Chores**
  * Migrated build/dependency management from Poetry to a setuptools + Pixi workflow.
  * Updated CI to use Pixi for formatting, linting, and testing (including editable install for discovery and updated coverage reporting).
  * Refreshed release automation with the latest release-please action and streamlined PyPI publishing.
  * Switched code style tooling to Ruff and updated related configuration.
  * Loosened the minimum version constraint for the `throttler` dependency.
* **Bug Fixes**
  * Improved the error message when array job submission is effectively disabled due to low array limits.


setup.py.bck
```diff
@@ -0,0 +1,57 @@
...
+setup(
+    name=project.get("name", "snakemake-executor-plugin-slurm"),
+    version=project.get("version", "2.1.0"),
+    description=project.get("description", ""),
+    author=", ".join(project.get("authors", [])),
+    license=project.get("license", ""),
+    url=project.get("repository", ""),
+    py_modules=[],
+    ext_modules=cythonize(extensions, language_level=3, compiler_directives={"linetrace": True}),
+    install_requires=list(dependencies.keys()),
```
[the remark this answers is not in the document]
cmeesters: @coderabbitai: This setup.py.bck was experimental and was accidentally pushed. It is deleted, not.

setup.py.bck
```diff
@@ -0,0 +1,57 @@
...
+        file_path = os.path.join(base_dir, filename)
+        py_files.append(file_path)
+
+# Create a single Extension with all Python files
+extensions = [
+    Extension(
+        'snakemake_executor_plugin_slurm._core',
+        py_files
+    )
+]
```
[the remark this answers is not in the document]
cmeesters: @coderabbitai: This setup.py.bck was experimental and was accidentally pushed. It is deleted, not.

.github/workflows/ci.yml
```diff
@@ -11,41 +11,25 @@ jobs:
...
-          python-version: "3.11"
-
-      - name: Install poetry
-        run: pip install poetry
-
-      - name: Install dependencies
-        run: |
-          poetry install --sync
+      - name: Setup pixi
+        uses: prefix-dev/setup-pixi@v0
```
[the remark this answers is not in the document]
johanneskoester: No, this is overkill. The pixi action has been very stable.

.github/workflows/release-please.yml
```diff
@@ -8,43 +8,32 @@ name: release-please
...
-      release_created: ${{ steps.release.outputs.release_created }}
     steps:
-      - uses: googleapis/release-please-action@v4
+      - uses: googleapis/release-please-action@v5
         id: release
         with:
           release-type: python
           token: ${{ secrets.RELEASE_PLEASE_PR_CI_TOKEN }}
 
-  publish:
```
johanneskoester, line 20: The separation into a release-please job and a publish job was actually a good thing. I would suggest to keep that, while still switching to pixi for the publish job.

pixi.toml
```diff
@@ -0,0 +1,30 @@
+[project]
```
johanneskoester: There is no need for a spearate pixi.toml. Pixi works fine with a pyproject.toml (see the docs). Pixi specific sections are just added as `[tool.pixi.*]`.

johanneskoester (changes requested):

pyproject.toml
```diff
@@ -1,49 +1,107 @@
...
+platforms = ["linux-64"]
+
+[tool.pixi.dependencies]
+python = ">=3.11,<3.14"
+pip = ">=24,<26"
+packaging = ">=24,<26"
+poetry = ">=2.4.1" # for local use
+
+[tool.pixi.tasks]
+install = { cmd = "poetry install" }
```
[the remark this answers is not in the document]
cmeesters: @coderabbitai I want an installation command which installs the plugin into my local conda env. So far, I cannot do this with pixi. Hints?

pixi.toml
cmeesters: The file pixi.toml was there, because I just started to understand pixi - and thought I need an additional file, if I do want to install dev versions into conda envs. (For the time being, I want to stay working with conda envs). However, this is unnecessary. Hence, the file is deleted.

.github/workflows/release-please.yml
```diff
@@ -8,43 +8,34 @@ name: release-please
...
-      - name: Install poetry
-        run: pip install poetry
-
-      - name: Determine dependencies
-        run: poetry lock
-
-      - uses: actions/setup-python@v5
+      - name: Set up Pixi
+        if: ${{ steps.release.outputs.release_created }}
+        uses: prefix-dev/setup-pixi@v0.9.6
```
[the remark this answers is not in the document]
cmeesters: @coderabbitai: I do want to pin to tags - pinning to commit SHAs makes updating harder as hashes need to be found and updating is error-prone.

.github/workflows/release-please.yml
cmeesters: yes. I reverted that change. This PR should focus on the CI, only. However, switching to pixi in for release-please might require some help: I have been tinkering a looong time to get the CI running as I was used to. Repeating this with release-please is not advisable as releases should be working without irritating users.

pyproject.toml
```diff
@@ -1,49 +1,119 @@
...
+[project.urls]
+Repository    = "https://github.com/snakemake/snakemake-executor-plugin-slurm"
+Documentation = "https://snakemake.github.io/snakemake-plugin-catalog/plugins/executor/slurm.html"
 
-[tool.poetry.scripts]
+[project.scripts]
 generate-slurm-partition-config = "snakemake_executor_plugin_slurm.cli:main"
 
+[python-entrypoint-group."snakemake executor plugins"]
+slurm = "snakemake_executor_plugin_slurm:ExecutorPlugin"
```
johanneskoester: What is this supposed to do??

pyproject.toml
```diff
@@ -1,49 +1,119 @@
...
+[tool.codespell]
+skip = ".git,*.pdf,*.svg,versioneer.py,*.css,test_*"
+check-hidden = true
+
 [tool.coverage.run]
 omit = [".*", "*/site-packages/*", "Snakefile"]
 
-[tool.black]
+[tool.ruff]
 line-length = 88
```
johanneskoester: This line can be removed now. It was there to make black interact fine with flake8. But with ruff, the defaults for formatting and linting already match.

```suggestion
```

pyproject.toml
```diff
@@ -1,49 +1,119 @@
...
-extend-exclude = 'tests/test_scontrol_parsing\.py'
+target-version = "py311"
+extend-exclude = ["tests/test_scontrol_parsing.py"]
+
+[tool.ruff.lint]
+ignore = ["E203"]
```
johanneskoester: I would definitely add a comment that explains why.

pyproject.toml
```diff
@@ -1,49 +1,119 @@
...
 
-[build-system]
-requires = ["poetry-core"]
-build-backend = "poetry.core.masonry.api"
+
+[tool.pixi]
+
+# Workspace table (required for preview features)
+[tool.pixi.workspace]
+preview = ["pixi-build"]
```
johanneskoester: Why do you want to have pixi-build here?

pyproject.toml
```diff
@@ -1,49 +1,119 @@
...
+channels = ["conda-forge", "bioconda"]
+platforms = ["linux-64"]
+
+# Conda-only dependencies (must be a table)
+[tool.pixi.dependencies]
+python    = ">=3.11,<3.14"
+pip       = ">=24,<26"
+packaging = ">=24,<26"
```
johanneskoester: Why do you need these dependencies?

pyproject.toml
```diff
@@ -1,49 +1,119 @@
...
+
+[tool.pixi.feature.test.pypi-dependencies]
+pytest    = ">=9.0.3,<10"
+coverage  = ">=7.14.0,<8"
+snakemake = ">=9.21.0,<10"
+
+[tool.setuptools.packages.find]
+include = ["snakemake_executor_plugin_slurm*"]
+# Exclude the `pr_logo` directory (and any sub‑files) from the packaged source
+exclude = ["pr_logo*"]
```
johanneskoester: Why is setuptools still needed now what you have migrated to pixi?

pyproject.toml
```diff
@@ -1,49 +1,119 @@
...
+
+[tool.pixi.feature.test.tasks]
+test            = {cmd = "pytest"}
+coverage-report = {cmd = "coverage run -m pytest -vv --capture=no && coverage report -m"}
+
+[tool.pixi.feature.style.pypi-dependencies]
+black = ">=24.2,<25"
```
johanneskoester: Why is black still needed?

pyproject.toml
johanneskoester: Ideally, avoid conda dependencies and remove the entire section.
The python version should instead be defined via the requires-python key under `[project]`.

.github/workflows/release-please.yml
johanneskoester: I don't see the revert. The new release please still has just a single job. It should hovever have two jobs as it had before. One running release-please, and a second one, depending on the first, that publishes the package after release-please has been executed.

johanneskoester (changes requested):

pyproject.toml
cmeesters: my ignorance ...

pyproject.toml
cmeesters: will try. Might have to do with my mingling with conda and pixi. Meanwhile, I have  pixi-only env, too.

pyproject.toml
cmeesters: hm, I tried without and got errors. Again, might be a leftover. Will test.

cmeesters:
@johanneskoester now, the revert should be pushed.

pyproject.toml
cmeesters: I _think_ this came from older code I took as a template. During experimenting, I did not remove it - it is removed now.

pyproject.toml
cmeesters: done

pyproject.toml
cmeesters: I had troubles installing the plugin locally. I _thought_ this was the fix. Turns out: No.

pyproject.toml
cmeesters: ah, I remember: I tried installing locally into a conda env with pixi and failed. So, I switched to pip. Keeping the spec in the toml lets me do that. It is not strictly necessary for pixi in the CI, yet for mixing Conda and pixi it is convenient.

I am open to alternatives: As of now, I am still new to pixi (and conda, infact).

pyproject.toml
cmeesters: a leftover from testing and comparisons. Gone.

pyproject.toml
cmeesters: see the other comment: Locally, not within the CI, I still seem to need pip to install in working conda envs. Yes, pixi allows for different envs. If we want to rely on pixi, only, we need to make a few adjustments, e.g. setting the pixi cache - which seems(!) impossible in the `pyproject.toml` (so, such settings cannot be shipped, if I am correct?).

pyproject.toml
cmeesters: I am making a statement out of ignorance and because I failed so far. So, is this statement correct: If we want pixi-only dependency resolvement for Snakemake. Snakemake and its plugings need to share a common environment defined in all pyproject.tomls?

pyproject.toml
johanneskoester: I would we need to set the pixi cache?
You just do `pixi shell -e dev` and then you are in the development env that the pyproject.toml defines.

pyproject.toml
johanneskoester: > Snakemake and its plugings need to share a common environment defined in all pyproject.tomls?

No. pixi will install the required dependencies, either from pypi, or if specified as such in the pyproject.toml, from local paths (also in editable mode if specified like that, similar to pip -e).

johanneskoester (approved):

Issue #353, opened 2025-09-10
no-requeue


conchoecia:
**Is your feature request related to a problem? Please describe.**

Sometimes my snakemake rules have a slow ramp-up on RAM usage, and upon failure the pipeline grants the rules more ram with the retry system. However, my SLURM env (and others) have requeue options that cause failed jobs to re-run a default of 5 times before the job is killed and retry has a chance to resubmit with more resources.

Effectively, this causes jobs to be run 4 extra times with failed parameters before snakemake with the slurm executor plugin can do its thing and resubmit the job again with more resources.

**Describe the solution you'd like**

I would like there to be a --no-requeue option for the slurm executor plugin to override the system or user defaults.

**Describe alternatives you've considered**

This behavior can work with the generic cluster executor for snakemake by passing the `--no-requeue` optional command. This could also work by changing user configs for SLURM, but that isn't very portable and could clobber someone's settings when I just want this behavior for some scripts. I don't know how this could be set up with the snakemake profile files, even if it could though that may be bit heavy-handed since there are strict requirements on the path of the snakemake profile files, reducing its portability.

From what I can tell, the options for requeue in the plugin are limited to a flag that turns requeue on (`--slurm-requeue`).

**Additional context**

Happy to submit a PR if you have ideas on what behavior you think would mesh well with the plugin!

gmanthey:
Hi all,

I've run into a related problem that would require a very similar fix. On my cluster requeuing is enabled by default. With no further parameters passed to snakemake, this in some cases causes snakemake to see a job running into a `NODE_FAIL`, stopping the pipeline, while slurm at the same time is requeuing the job, which then runs through normally, leading to a state where snakemake reports a job as failed, but logs and slurm output suggests the job ran through just fine. Adding retrying in snakemake then leads to a state where 2 jobs run at the same time and we even had it happen that the pipeline failed because the node that snakemake tried to mark as failed was "None assigned", as slurm had already requeued the failed job at the time that snakemake queried to get the failed node.

For us we found that we'd either need to pass `--no-requeue` to slurm or explicitly tell the slurm plugin to requeue using `--slurm-requeue`, but I don't think this should be up to the user to figure out.

I see 2 solutions to this: 
1. By default detect if requeuing is enabled by default on the cluster and act accordingly. Add the options `--slurm-requeue=true` and `--slurm-requeue=false` to explicitly set requeuing or no requeuing.
[...]

PR #476, opened 2026-07-15
feat: add no-requeue option for #353

conchoecia:
Addresses my comments in https://github.com/snakemake/snakemake-executor-plugin-slurm/issues/353#issue-3402907753. Simply adds a `--no-requeue` option to prevent SLURM from re-queueing jobs, thereby allowing snakemake to resubmit the jobs with the proper hooks that allow for increased resources on subsequent submissions.

## Summary by CodeRabbit

* **New Features**
  * Added an option to disable SLURM’s automatic job requeue behavior.
  * Disabled requeueing is applied through generated SLURM submission commands.
  * Prevented simultaneous use of requeue and no-requeue options.

* **Documentation**
  * Documented the new configuration option and its interaction with retry behavior.
  * Added the corresponding SLURM resource mapping and command-line guidance.

### DIFF
--- a/docs/further.md
+++ b/docs/further.md
@@ -227,6 +227,7 @@ These are the available options, and the SLURM `sbatch` command line arguments t
 |                      | passed through when explicitly set) |                     |
 | `slurm_account`      | account for resource usage tracking | `--account`         |
 | `slurm_partition`    | partition/queue to submit job(s) to | `--partition`       |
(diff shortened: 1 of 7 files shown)


cmeesters (approved):
Thank you for this PR.

I like the idea of having more explicit settings and dislike the idea of cluttering the help output and increasing the complexity.

Sure hope, we can soon hide options which are for internal use, only.

PR #477, opened 2026-07-20
feat: Allow disabling array memory fudge

lewis-k-s:
## Summary

- add an executor setting that controls array memory fudge application
- preserve the existing behavior by default
- allow profiles to disable the synthetic memory request on clusters where memory is derived from CPU allocation

## Motivation

Array submission currently calls `apply_mem_fudge()` unconditionally. If a job has no explicit memory resource, that helper adds `--mem 1`. Some Slurm clusters derive host memory from requested CPUs and reject or misinterpret any explicit memory option, so native arrays cannot be used there safely.

The new `array_memory_fudge` setting defaults to true. Existing users therefore retain the current behavior, while affected profiles can set:

```yaml
slurm-array-memory-fudge: false
```

## Validation

- existing array-memory behavior remains covered with the default enabled
- disabling the setting emits neither `--mem` nor `--mem-per-cpu`
- profile-style true/false values are parsed explicitly
- focused array executor suite passes (47 tests at this commit)

### DIFF
--- a/docs/further.md
+++ b/docs/further.md
@@ -345,6 +345,22 @@ When submitting array jobs, the `--slurm-array-limit` flag defines the
 maximum number of array tasks to be submitted in one job submission.
 If the number of tasks exceeds this limit, multiple array job submissions will be performed. This is useful to avoid hitting cluster limits on the maximum number of array tasks per job. Please obey your cluster limits and set this flag accordingly.
 
+##### Array memory adjustment
+
+By default, the plugin increases an explicit memory request for an array job to
+account for the encoded array-job payload. If a job has no memory constraint,
+the plugin adds a minimal `--mem` request so that this adjustment is not lost.
+
+Some clusters derive memory allocation from other requested resources and do
+not allow an explicit memory option. Disable the adjustment on such clusters:
+
+```console
+snakemake --slurm-array-memory-fudge false ...
+```
+
+The default is `true`, preserving the standard array submission behavior. In a
+Snakemake profile, use `slurm-array-memory-fudge: false` instead.
+
 
 #### MPI-specific Resources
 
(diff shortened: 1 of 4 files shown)


cmeesters:
@lewis-k-s Thank you for this PR!

I will merge it, as soon, as it is not a draft PR any more:

- As for the formatting: There is another PR dealing with a switch to flake8/ruff, so, when this PR is merged, I will do the formatting again.
- There is this little boolean conversion. Currently, I'm on holiday and changing positions. I might want to check, whether this can be accomplished differently - Snakemake will get a new CI in the not-so-near future and I do not want code redundancy.

Apart from these two minor issues, there is only the missing documentation: Do you want to write it? Else, I will do it. I found a number of documentation issues and an overhaul is due. All this (particularly me getting started elsewhere) might take a few weeks until a release, though. I try to be faster, but cannot make guarantees.

lewis-k-s:
Hi @cmeesters :)

I’ve added the documentation. I agree the parser could become shared infrastructure; until that lands, this local explicit converter avoids Python’s truthiness trap. I also added a settings-resolution regression test to verify the behavior end-to-end. I won't mark it ready yet because maintainer intervention is necessary for the [CI workflow](https://github.com/snakemake/snakemake-executor-plugin-slurm/actions/runs/30630782956)

cmeesters:
Going on holiday helps ...

- I think, I will need to implement the pooling of smp jobs into one or multi-node jobs, a.s.a.p. For this we will need the same of a more sophistically calculated memory overhead. If one wants to turn this off, we could use the same flag. Then, however, the naming is not good. What if we drop the `_array` string in this flag?
- Also, the boolean flags so far, simply have a default (in this case the turn-off default could be `False`), and the flag become `--slurm-disable-memory-fudge`.

What do you think? Do you want to do it? Shall I? It's not much work. In any case this is then not a draft PR anymore.

cmeesters:
@lewis-k-s closing this PR in favour of #482 . As of v2.8.0 (released today, so might take a while to be in Conda) `--slurm-disable-memory-fudge` is present. I'm afraid, this is a breaking change compared to your current code base, as I have to assume. But due to upcoming changes, your PR would introduce a breaking change for all users in the future. 

Thanks again for your contribution!

